PR #27, opened 2019-03-18
Fixed #25

### LINKED ISSUE #25: windows ml-api build fail
I was able to start the build under windows in powershell by cd to the TSD directory and then the normal docker-compose up -d, I'm sure there is a syntax that would make it one line but I'm not that versed.

The problem I get is:

ERROR: Service 'ml_api' failed to build: The command '/bin/sh -c wget --quiet -O model/model.weights $(cat model/model.weights.url)' returned a non-zero code: 8

Any ideas? I checked the url in the file and was able to download it manually in a browser so it's reachable from my pc

goopsie:
Fixes #25

Don't have a linux or mac machine to test on, but should work there as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ git clone https://github.com/TheSpaghettiDetective/TheSpaghettiDetective.git
 2. Run it! Do **either** one of these based on what OS you are using:
     - If you are on Linux: `cd TheSpaghettiDetective && sudo docker-compose up -d`
     - If you are on Mac: `cd TheSpaghettiDetective && docker-compose up -d`
-    - If you are on Windows: ??? (help needed here)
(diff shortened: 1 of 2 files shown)


kennethjiang:
@goopsie thank you for helping us with Windows platform. It's arguably the most important platform as I bet most people would run the server on a Windows. It's just the TSD team is Mac users so Windows has been a bit unduly neglected. lol.

For your PR, I'm a bit hesitant to put the model URL into `Dockerfile`. Model file will change regularly as we keep on improving it, but we hope we can keep `Dockerfile` stable to have a clean history for it (hopefully only changes related to the docker images are in its history).

Plus I i'm curious why `$(cat ...)` would fail in a container on Windows. That command should be running inside a Linux container and it should be a valid syntax.

I'll try to spin up a Windows VM to test it out. In the meanwhile I'll keep your PR open.

PR #55, opened 2019-04-20
When building an image with circle ci tag that image as the latest image

wwsean08:
It's pretty common to have the latest tag rolling with the newest image, this change updates the circle ci process to tag it's current build as latest.


kennethjiang:
Thanks @wwsean08 for your contribution!

PR #56, opened 2019-04-21
Reduce docker image sizes and update docker compose to handle alpine images

### LINKED ISSUE #42: Suggestions for a faster build process
Hi, I just stumbled upon your project, it looks really cool.
In the readme you mentioned the long build process. I just skimmed over your docker configuration and have a few suggestions:

- [ ] **ml_api**: Use `nvidia/cuda:9.0-runtime-ubuntu16.04` instead of `nvidia/cuda:9.0-devel-ubuntu16.04`, it should be sufficient and uses 1GB less space. -> Saves download time
- [ ] **web**: Use `python:3.6-alpine` instead of `python:3.6` it uses 850MB less space, vim and ffmpeg are available in alpine, vim can probably be removed. -> Saves download time
- [ ] **web**: Make better use of docker caching: First add the requirements file, then pip install, then add the rest of the project. Using this the build process is much faster if no package requirements have changed. -> Faster development cycle
- [ ] **redis** (minor): Use `redis:alpine`, uses 50 MB less space
- [ ] **all docker images**: build them and upload them to Dockerhub. Most of the time the build process downloads and installs apt and pip packages. By prebuilding and uploading to Dockerhub, this can be consolidated into one docker pull for each image. If you still provide the Dockerfiles, anybody can build the images themselves when they want to debug something. -> Saves build time.

By using all suggestions above, you could get rid of all package installations (using the prebuilt image) which saves a lot of time and also save 1.9GB of Downloads.

wwsean08:
fixes #42

# Overview of the changes
* ml_api now uses 9.0-runtime-ubuntu16.04 insead of 9.0-devel-ubuntu16.04
* ml_api no longer installs recommended packages (which are not requirements) and explicitly add a few packages because of this
* web now is based off of alpine
* combined a few run statements to reduce the layer count
* rearranged statements so things that don't change often (like requirements.txt or the model weights) are cached on rebuilds
* redis is now using the 5.0-alpine image

I've done some basic testing validating that the user interface seems to work as well as spaghetti detection by adding some to my print bed mid print manually, however I would not call the testing exhaustive, if there are any sort of regression tests I can run thru this I'd love to know so I could feel more confident about this change.

Image of a detected failure with the other docker image:
image

# Benchmarks:
## Machine Specs
This testing was done in a VirtualBox VM with Ubuntu 18.04 with 16GB of memory, and 4 cores (2 logical, 2 hyperthreads).

## Speed
Take note of the second test where there is a minor change to the code, that is where this change truly shines in terms of speed.

docker-compose up original (no images cached): 
> time docker-compose up -d
> ...... lots of stdout
> Creating thespaghettidetective_ml_api_1 ... done
> Creating thespaghettidetective_redis_1  ... done
> Creating thespaghettidetective_web_1    ... done
> Creating thespaghettidetective_tasks_1  ... done
> real	10m0.465s
> user	0m2.017s
> sys	0m0.370s

docker-compose up new (no images cached): 
> time docker-compose up -d
> ...... lots of stdout
> Creating thespaghettidetective_ml_api_1 ... done
> Creating thespaghettidetective_redis_1  ... done
> Creating thespaghettidetective_web_1    ... done
> Creating thespaghettidetective_tasks_1  ... done
> real	8m1.155s
> user	0m1.440s
> sys	0m0.300s

docker-compose up --build original (with a minor code change to the web project)
> time docker-compose up -d --build
> ...... lots of stdout
> Creating thespaghettidetective_redis_1  ... done
> Creating thespaghettidetective_ml_api_1 ... done
> Creating thespaghettidetective_web_1    ... done
> Creating thespaghettidetective_tasks_1  ... done
> real	**1m57.467s**
> user	0m1.069s
> sys	0m0.248s

docker-compose up --build new (with a minor code change to the web project)
> time docker-compose up -d --build
> ...... lots of stdout
> Creating thespaghettidetective_redis_1  ... done
> Creating thespaghettidetective_ml_api_1 ... done
> Creating thespaghettidetective_web_1    ... done
> Creating thespaghettidetective_tasks_1  ... done
> real	**0m15.935s**
> user	0m0.932s
> sys	0m0.266s

## Image Sizes
ml image original: 2.95GB
ml image new: 1.67GB
web image original: 1.39GB
web image new:  504MB

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -2,7 +2,8 @@ version: '2.4'
 
 x-web-defaults: &web-defaults
   restart: unless-stopped
-  image: thespaghettidetective/web:0.123
+  build:
+    context: web
   volumes:
     - ./web:/app
   depends_on:
@@ -34,7 +35,8 @@ services:
   ml_api:
     hostname: ml_api
     restart: unless-stopped
-    image: thespaghettidetective/ml_api:0.123
+    build:
+      context: ml_api
     environment:
         DEBUG: 'True'
         FLASK_APP: 'server.py'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,13 +51,13 @@ services:
       - "3334:3334"
     depends_on:
       - ml_api
-    command: bash -c "python manage.py collectstatic --noinput && python manage.py migrate && python manage.py runserver 0.0.0.0:3334"
+    command: sh -c "python manage.py collectstatic --noinput && python manage.py migrate && python manage.py runserver 0.0.0.0:3334"
 
   tasks:
     <<: *web-defaults
     hostname: tasks
-    command: bash -c "celery -A config worker -l info"
+    command: sh -c "celery -A config worker -l info"
 
   redis:
     restart: unless-stopped
-    image: redis:5.0
+    image: redis:5.0-alpine
--- a/ml_api/Dockerfile
+++ b/ml_api/Dockerfile
@@ -1,5 +1,6 @@
 FROM nvidia/cuda:9.0-runtime-ubuntu16.04
 WORKDIR /app
+EXPOSE 3333
 RUN apt update && \
     apt install --no-install-recommends -y libsm6 libxrender1 libfontconfig1 python3-pip python3-setuptools vim ffmpeg wget curl && \
     rm -rf /var/lib/apt/lists/* && \
@@ -14,4 +15,3 @@ RUN wget --quiet -O model/model.weights $(cat model/model.weights.url | tr -d '\
 
 ADD . .
 
-EXPOSE 3333
--- a/ml_api/Dockerfile
+++ b/ml_api/Dockerfile
@@ -1,14 +1,17 @@
-FROM nvidia/cuda:9.0-devel-ubuntu16.04
-
-RUN apt update
-RUN apt install -y libsm6 libxrender1 libfontconfig1 python3-pip vim ffmpeg wget curl
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
-RUN update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 10
-
-ADD . /app
+FROM nvidia/cuda:9.0-runtime-ubuntu16.04
 WORKDIR /app
-RUN wget --quiet -O model/model.weights $(cat model/model.weights.url | tr -d '\r')
+RUN apt update && \
+    apt install --no-install-recommends -y libsm6 libxrender1 libfontconfig1 python3-pip python3-setuptools vim ffmpeg wget curl && \
+    rm -rf /var/lib/apt/lists/* && \
+    update-alternatives --install /usr/bin/python python /usr/bin/python3 10 && \
+    update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 10
+
+ADD requirements.txt .
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
+ADD model model
+RUN wget --quiet -O model/model.weights $(cat model/model.weights.url | tr -d '\r')
+
+ADD . .
 
 EXPOSE 3333
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -1,11 +1,15 @@
-FROM python:3.6
+FROM python:3.6-alpine3.9
 
-RUN apt-get update && apt-get install -y vim ffmpeg
+WORKDIR /app
+RUN apk -U add vim ffmpeg postgresql-libs && \
+    apk add --virtual .build-deps gcc musl-dev postgresql-dev zlib-dev jpeg-dev
 RUN pip install --upgrade pip
 
-ADD . /app
-WORKDIR /app
-RUN pip install -r requirements.txt
+ADD requirements.txt .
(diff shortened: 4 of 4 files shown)


kennethjiang:
Awesome! I like this PR @wwsean08 

I'll test this PR tomorrow when I get a chance.

I'm under the impression that most of the build time is spent on installing the system packages and python pips. Have you tried saving an image with these packages installed already, and change `FROM` to using that image? I suspect that would significantly reduce the time to run `docker-compose up -d` for the first time.

wwsean08:
Having a base image that has all those apt and pip installs (plus downloads the model weights) would definitely be a significant reduction, alternatively there could be 2 docker compose files, a development one like what currently exists and a production one that points to static images on dockerhub.

My personal preference coming from using kubernetes is to have the images well defined and tagged and not build them at startup from a deployment scenario, only doing that in a development scenario.

wwsean08:
For reference I just tested on an underpowered vm on a laptop and with a slightly modified docker compose file that pointed at the images currently on dockerhub it took just under 4 minutes to startup almost all of that was download/extraction time which the changes outlined above would make even faster given how much the images got trimmed down.  I'm adding another commit which basically makes it so there are two docker compose files, one for dev and one for deployment (that being the default docker-compose.yml)  If you disagree with this idea I can back it out.  The downside of this method is that releases need to be thought out as the docker compose file needs to be updated to point at the new image tag.

kennethjiang:
K will test it in our staging asap

wwsean08:
Sounds good, just as a reminder to test my latest changes you'll need to run `docker-compose -f docker-compose-dev.yml up` as this has my changes as opposed to running `docker-compose up` which will just pull version 0.123 from dockerhub (instead of building them in place like they used to) for the various images.

kennethjiang:
I got
```
ERROR: Service 'web' failed to build: The command '/bin/sh -c pip install -r requirements.txt &&     apk --purge del .build-deps' returned a non-zero code: 1
```
When i shelled into the offending container and ran that command, I got:
```
/app # /bin/sh -c pip install -r requirements.txt

Usage:
  pip <command> [options]

Commands:
  install                     Install packages.
  download                    Download packages.
  uninstall                   Uninstall packages.
  freeze                      Output installed packages in requirements format.
  list                        List installed packages.
  show                        Show information about installed packages.
  check                       Verify installed packages have compatible dependencies.
  config                      Manage local and global configuration.
  search                      Search PyPI for packages.
  wheel                       Build wheels from your requirements.
  hash                        Compute hashes of package archives.
  completion                  A helper command used for command completion.
  help                        Show help for commands.

General Options:
  -h, --help                  Show help.
  --isolated                  Run pip in an isolated mode, ignoring environment variables and user configuration.
  -v, --verbose               Give more output. Option is additive, and can be used up to 3 times.
  -V, --version               Show version and exit.
  -q, --quiet                 Give less output. Option is additive, and can be used up to 3 times (corresponding to WARNING, ERROR, and CRITICAL logging levels).
  --log <path>                Path to a verbose appending log.
  --proxy <proxy>             Specify a proxy in the form [user:passwd@]proxy.server:port.
  --retries <retries>         Maximum number of retries each connection should attempt (default 5 times).
  --timeout <sec>             Set the socket timeout (default 15 seconds).
  --exists-action <action>    Default action when a path already exists: (s)witch, (i)gnore, (w)ipe, (b)ackup, (a)bort).
  --trusted-host <hostname>   Mark this host as trusted, even though it does not have valid or any HTTPS.
  --cert <path>               Path to alternate CA bundle.
  --client-cert <path>        Path to SSL client certificate, a single file containing the private key and the certificate in PEM format.
  --cache-dir <dir>           Store the cache data in <dir>.
  --no-cache-dir              Disable the cache.
  --disable-pip-version-check
                              Don't periodically check PyPI to determine whether a new version of pip is available for download. Implied with --no-index.
  --no-color                  Suppress colored output
```

It seems that pip in python-3.6-alpine doesn't take `-r` param. Do you know where I did it wrong @wwsean08  ?

wwsean08:
Hmm I don't run into that issue, see that the -r parameter does exist, whats the sha digest for the python-3.6-alpine you're using that way I can make sure I'm testing against the exact same image.  To get that simply run `docker inspect python:3.6-alpine3.9 --format='{{index .RepoDigests 0}}'`

kennethjiang:
It turns out that it was `&&     apk --purge del .build-deps` that was causing the failure. Didn't want to figure out why so I just deleted that line. Hopefully it doesn't cost too much more in image size.

Also I modified your PR a bit before merging. Feel free to take a look and make further suggestions. @wwsean08 . I tested it on a blank VM in the cloud (so image download was very fast) and it took 1:32 to finish `docker-compose up`. :)

wwsean08:
Strange the apk purge wasn't working, it's not a huge difference, just a couple meg.

Ok I see the modifications you made and it looks good, the one suggestion I would make is that I would version the base image because if for example a dependency is added it's not a huge deal but if you upgrade or remove a dependency then it would be bad for someone to accidentally pull the wrong base image, so maybe something like base-1.0.0, then increment the semver as needed.  I'm just always concerned about reproducability with shared tags, makes debugging easier :P

wwsean08:
Actually now that I look closer it's not a huge deal

PR #57, opened 2019-04-30
SMS Notification and Home button

Phat32:
The current SMS Notification uses a hardcoded URL instead of the configured URL. This just replaces the hard-coded URL with the generated one.

The Home button is hardcoded for the home page of the main site which is already accessible by clicking the logo to the left of Home. This just navigates to / when clicking Home for easier navigation.

### DIFF
--- a/web/app/templates/layout.html
+++ b/web/app/templates/layout.html
@@ -58,7 +58,7 @@
       <div class="collapse navbar-collapse" id="navbarResponsive">
         <ul class="navbar-nav">
           <li class="nav-item">
-            <a class="nav-link" href="https://www.thespaghettidetective.com/">Home</a>
+            <a class="nav-link" href="/">Home</a>
           </li>
           <li class="nav-item">
             <a class="nav-link" href="https://www.thespaghettidetective.com/help">Help</a>
--- a/web/app/templates/layout.html
+++ b/web/app/templates/layout.html
@@ -58,7 +58,7 @@
       <div class="collapse navbar-collapse" id="navbarResponsive">
         <ul class="navbar-nav">
           <li class="nav-item">
-            <a class="nav-link" href="/">Home</a>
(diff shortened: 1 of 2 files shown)


web/app/templates/layout.html
```diff
@@ -58,7 +58,7 @@
       <div class="collapse navbar-collapse" id="navbarResponsive">
         <ul class="navbar-nav">
           <li class="nav-item">
-            <a class="nav-link" href="https://www.thespaghettidetective.com/">Home</a>
+            <a class="nav-link" href="/">Home</a>
```
kennethjiang: `/` is actually `https://app.thespaghettidetective.com`, will actually be redirected to "My Printers" page.  `https://www.thespaghettidetective.com/` is the static html page that gives user general info about TSD.

Phat32: `https://www.thespaghettidetective.com/` is also accessible from the logo located to the left of the Home button. When using the interface and viewing the gallery, I will press "Home" to get back to printers and personally will get a little annoyed that it goes to another URL entirely. The `My Printers` link can be clicked, but default behaviour drilled into us is to click the `Home` link when available.

kennethjiang: It's a good point. Our original idea was to have the links at the left hand side to go to static content: web site, help docs, etc. And the links at the right hand side to go to the app itself. I'm open to having Home to point to "My Printers" (in this case we will want to just keep one of them), but then we probably need to re-organize the nav links in a way not to confuse users (why some links are on the left while others on the right).

Phat32:
Removed the Layout page changed leaving just the SMS notification changes

kennethjiang:
@Phat32 thanks again for this PR. we have been tied up with other priorities. will consider your suggestion on changing the nav bar when we get a chance

PR #65, opened 2019-05-14
Fix spelling/grammar

goopsie:
Not that useful, but makes it look a bit nicer


kennethjiang:
Haha never realized we made so many grammatical and spelling errors. One of the best PRs we have received so far! :)

PR #69, opened 2019-05-29
Fixes #68 making the home button go to the home of the instance you are using

### LINKED ISSUE #68: Home link should go to running instance
Currently "home" on the upper left always goes to https://www.thespaghettidetective.com even if you are running your own.  It should be updated to the running instance.  Preferably the logo also would but I can understand if you want that to go to the central site.

wwsean08:
This change outlines what I explained in #68, setting the home button to / which will always be the home of the server you're connected to whether it's The Spaghetti Detective or whether you're running it yourself.

### DIFF
--- a/web/app/templates/layout.html
+++ b/web/app/templates/layout.html
@@ -58,7 +58,7 @@
       <div class="collapse navbar-collapse" id="navbarResponsive">
         <ul class="navbar-nav">
           <li class="nav-item">
-            <a class="nav-link" href="https://www.thespaghettidetective.com/">Home</a>
+            <a class="nav-link" href="/">Home</a>
           </li>
           <li class="nav-item">
             <a class="nav-link" href="https://www.thespaghettidetective.com/help">Help</a>


kennethjiang:
This is an issue we have back-burnered - what's the optimal way to arrange the nav links.

Our current thinking is that links on the left-hand side point to the static/pubic website, whereas the links on the right-hand side point to user's own pages. The rationale behind it was to provide a way for users to jump to important pages relatively easily. But we are open to suggestions.

It does make sense to point "Home" to the current instance. However, we need to solve 2 problems that come with it:

1. `/` is actually the same as `/printers` because of internal redirect. So these 2 links: "Home" and "My Printers" point to the same page, which is waste of precious nav bar real estate and quite confusing.

2. The other 2 links on the left-hand side point to the public site. It creates some inconsistency to have a link next to them going to user's own page.

Let me know what you think @wwsean08

wwsean08:
I'm fine with that though it's not entirely intuitive in terms of UX, my main thing is that when I look at home I think of the home page of the application, this was just something I noticed while playing around with some other stuff I was working on and seemed like an easy/quick MR/issue.

Maybe rewording it might make it a bit clearer for new users who try to set it up since normally something like a logo or "home" in the upper left does stay on the same server.

kennethjiang:
Agreed the nav bar is not intuitive. Have intended to get a UI/UX designer to look at it and make suggestion but it has not been a priority.

Any suggestion is welcome on how we should restructure the nav bar.

wwsean08:
I'll close this out, I may play with the UX a bit down the line as I'm not a UX expert by any means but am interested and do have a few ideas to play around with and run by some friends that know a bit more

kennethjiang:
Thank you Sean. And it'll be awesome if you can find help on UX/UI design. This the skill current TSD team doesn't have. :)

PR #72, opened 2019-05-31
Initial implementation to fix #63

### LINKED ISSUE #63: Send verification email and require verification for notification
TheSpaghettiDetective should send a verification email to users when they sign up with email/password to verify they actually own the account.  Social login proves they control the account and thus should control the email address associated with it so no need to send a verification in that case.

On top of that if an email address is not verified then it shouldn't get email notifications.  Most likely this will be a breaking change if it gets implemented fully (given the specification of don't notify if the emails not verified), so I'm open to feedback or alternative ideas, though it should be possible to write some code to send a verification email to every user at once as an admin (haven't validated that).

IMO this is a requirement for #46

wwsean08:
This change fixes #63 and makes the following changes:
* Email confirmations are sent when an email server is configured for new users (non-social)
  * Social accounts are verified by default (this isn't a change I made, just calling it out
* By default a user is required to confirm their email address to receive notification emails  **behavior change**
  * The EMAIL_REQUIRES_VERIFICATION allows you to override this behavior
  * It would be trivial to make the old behavior default
* All (verified) email accounts receive notification emails on print failure
* Displays the verification status for the primary email address in user preferences
* Add emails UI link to the user preferences where users can add email addresses to their accounts, change their primary email address, remove email accounts, and resend verification emails
* Skin the email confirmation/email UI
* If you have no verified email addresses confirmed and EMAIL_REQUIRES_VERIFICATION is true then a banner shows up in the UI for users to warn them they won't get notifications


web/app/templates/layout.html
```diff
@@ -103,9 +103,13 @@
   </nav>
   {% endwith %}
 
+  {% if settings.EMAIL_REQUIRES_VERIFICATION and user.is_authenticated and not user.has_verified_email %}
```
wwsean08: I know this if statement sucks but I couldn't come up with a better way of putting this notification in every page, if you have a suggestion, I'd love to hear it.  Also putting it outside of the div was intentional

wwsean08:
I have done minimal testing re: social accounts, this weekend I plan to do a more in depth testing to make sure that exposing this Email page won't break them in some way

kennethjiang:
Thank you @wwsean08 Let us know when you think this PR is ready.

web/app/templates/layout.html
kennethjiang: This `if` is clean and self-explanatory. Not a problem at all.

wwsean08:
Testing completed successfuly with social accounts, I was able to verify that adding email addresses, changing the primary email address, etc do not cause issues, and they act the same as normal accounts.  Please review whenever you have time, but everything seems good from my testing.  I also have committed making the default behavior of requiring accounts to be verified to receive an email false like it currently is.

kennethjiang:
Great. Will look at it sometime today.

kennethjiang:
Merged. Thank you so much @wwsean08 .

I made 2 major changes and a few minor UI changes to your PR. Let me know if you see problem with my changes:

1. removed `EMAIL_REQUIRES_VERIFICATION`. use `ACCOUNT_EMAIL_VERIFICATION` in place of it instead. the rationale was differences between seemed to be too subtle to be worth another configuration item. Instead, just assume that when `ACCOUNT_EMAIL_VERIFICATION` is not `none`, alerts won't be sent to unverified emails and hopefully this simplified logic is good for most, if not all, users.

2. Removed the javascript in `email.html`. Since it was very easy to add email back (kudos to you on this), the extra confirmation step doesn't seem to be worth it. Plus somehow it didn't work for me.

wwsean08:
That all sounds good, i'm not familiar with that javascript in question, didn't have any problems with it but also didn't notice it existed either.

kennethjiang:
Just rolled it out but noticed one thing: the confirmation email I received was sent in plain text version. It seems to be using this template: https://github.com/TheSpaghettiDetective/TheSpaghettiDetective/blob/master/web/app/templates/account/email/email_confirmation_message.txt . I remember it used to be html, which is this template: https://github.com/TheSpaghettiDetective/TheSpaghettiDetective/blob/master/web/app/templates/account/email_confirm.html

Any quick thought on why this changed @wwsean08 ?

wwsean08:
Not sure, all of my tests since enabling email confirmation always sent the text version of the message.  I can look into it, but my understanding was that template was for the actual webpage you use which is loaded from the text one, but I could be wrong.

kennethjiang:
Gotcha! I must have misremembered it. Text emails are ok. But I guess I will change the wording a little bit.

wwsean08:
Yup, just verified, I can create a template and open a PR if that's something you want, https://django-allauth.readthedocs.io/en/latest/advanced.html#sending-email

kennethjiang:
Nah. It was more about the fact that the template is using site name in the email subject and body. Previously the site name was `app.thespagehttidetective.com`. Now I changed it to `The Spagehtti Detective` and the text email is just fine. No need to go fancy. :)

wwsean08:
Ah ok, so something I couldn't know without looking at the actual admin console, got it :)

kennethjiang:
Sorry to go back and forth on this. Now I think it's still a good idea to implement an html version.

The reason is that some user's email client on mobile show the partial link, because of the `:` in the verification key. Screenshot attached.

When they clicked on that link, they actually got a 500 error. I was able to reproduce that by manually changing the key in the verification link.

I believe an html version of the verification email will solve the problem but let me know if you have a better idea. Thanks a lot @wwsean08 

Screenshot_20190604-132729

wwsean08:
Ok sounds good, I'll work on writing up an HTML version tomorrow and likely have a PR open for it either tomorrow night or Thursday

kennethjiang:
No worries @wwsean08 . I opened an issue #76 to keep track of it. I can pick it up myself when/if I get a chance.

Thanks again for this PR. Sending to only verified email addresses will greatly improve the reputation of all emails coming from `thespaghettidetective.com`

rubdos:
@kennethjiang, could you please do a decent job at censoring user data? This is certainly *not* within the very reasonable bounds of the GDPR.

Also note that some people don't read HTML mails, including myself.

wwsean08:
Once I have a template in place the email that gets sent will include both the HTML version and a text version to fall back on if a users client is configured to not allow html emails according to the documentation @rubdos, so this would work fine for you and other users that don't like/want html emails.

PR #87, opened 2019-08-16
Add Telegram bot to deliver telegram notifications

ajdavids:
This bot requires that you set up a bot with Telegram. See the updated
README.md to see how.

It also, albeit only briefly, requires that you have apublically-
accessible domain name.

There are two methods of using it - with webhooks, if TSD is running on
a domain accessible from the internet, and without, if it is not.

If you're using a local tunnel and want to manually specify the domain
TSD will run on, replace `telegram_bot.py:24` with
`bot.set_webhook('https://example.com/telegram/`)`, replacing
`https://example.com/` with your local tunnel URL. This will enable full
interaction with the bot even when running the server locally.


web/app/forms.py
```diff
@@ -47,3 +54,17 @@ def clean(self):
...
+
+        if bot and data['telegram_chat_id']:
+            auth = json.loads(data['telegram_chat_id'])
+            data['telegram_chat_id'] = auth['id']
+            telegram_chat_id = data['telegram_chat_id']
+
+            # When a user logs in, we save their chat id and their secret.
+            # We use their secret to validate web hooks,
+            # so we can query for the user and be sure they're not being spoofed.
+            data['telegram_secret'] = token_hex(12)
```
kennethjiang: Any reason why we need another secret here? Can we just use the secret token of the printer?

web/app/forms.py
```diff
@@ -47,3 +54,17 @@ def clean(self):
...
                 self.add_error('pushbullet_access_token',
                                'Invalid pushbullet access token.')
+
+        if not data['telegram_chat_id']:
+            data['telegram_chat_id'] = None
+
+        if bot and data['telegram_chat_id']:
+            auth = json.loads(data['telegram_chat_id'])
+            data['telegram_chat_id'] = auth['id']
+            telegram_chat_id = data['telegram_chat_id']
```
kennethjiang: I think these 2 lines can be simplified as `telegram_chat_id =  auth['id']`.

web/app/notifications.py
```diff
@@ -4,34 +4,63 @@
...
 import logging
 from urllib.parse import urlparse
 import ipaddress
 
 from app.models import Printer
+import app.telegram_bot
 from lib import site
 
 LOGGER = logging.getLogger(__name__)
+TELEGRAM_BOT = TeleBot(settings.TELEGRAM_BOT_TOKEN)
```
kennethjiang: good refactoring

web/app/templates/user_preferences.html
```diff
@@ -102,6 +102,16 @@ <h3>Pushbullet</h3>
             {% endif %}
           </div>
         </div>
+        <br />
+        {% if bot_name %}
+        <h3>Telegram</h3>
```
kennethjiang: any reason why we split telegram-related templates in 2 places? either we move this block all to `web/app/templates/telegram_button.html `, or we `web/app/templates/telegram_button.html ` to here.

web/app/urls.py
```diff
@@ -16,4 +16,5 @@
     path('prints/', views.prints, name='prints'),
     path('prints/delete/<pk>/', views.delete_prints, name='prints_delete'),
     path('user_credits/', views.user_credits, name='user_credits'),
+    path('telegram/', views.telegram, name='telegram'),
```
kennethjiang: Let's use `channels/telegram/` as the url.

web/app/views.py
```diff
@@ -4,21 +4,50 @@
...
+from ipaddress import ip_address, ip_network
+
 from .models import *
 from .forms import *
 from lib import redis
 from lib.channels import send_commands_to_group
+from .telegram_bot import bot_name, handle_callback_query
+
+# This list from https://core.telegram.org/bots/webhooks
+TELEGRAM_IPS = [ip_network('149.154.160.0/20'), ip_network('91.108.4.0/22')]
```
kennethjiang: Let's move everything related to telegram to `telegram_bot.py`

web/app/templates/telegram_button.html
```diff
@@ -0,0 +1,33 @@
...
+<script async src="https://telegram.org/js/telegram-widget.js?7"
+  data-telegram-login="{{ bot_name }}"
+  data-size="large"
+  data-onauth="onTelegramAuth(user)"
+  data-request-access="write"></script>
+{% endif %}
+<div class="col-sm-12">
+<div class="btn btn-primary float-left disabled"
+  style="display:none"
+  id="id_telegram_submit_prompt">Save to submit changes</div>
```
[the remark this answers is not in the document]
ajdavids: Sure, will make that change.

web/app/forms.py
ajdavids: 👍 Didn't know if you needed to explicitly set data attributes on a form for Django to save them out with those. Thanks.

web/app/templates/user_preferences.html
ajdavids: Got it. That was mostly for historical reasons. Will move all the telegram stuff in here.

web/app/forms.py
ajdavids: A user has a telegram login, not a printer. Can't a user have a bunch of printers? Or zero printers?

I could get rid of this secret if they're guaranteed to have at least one printer, but I don't think that's the case.

ajdavids:
How the telegram secret works overall is just as something temporary to let us verify callbacks. Instead of the DB, I can stick a random string in redis whenever a notification is sent and remove it whenever an action is taken. That would be more secure too, I think. I'll do that.

web/app/validators.py
```diff
@@ -0,0 +1,30 @@
...
+from time import time
+
+def validate_telegram_login(auth_dict):
+    if not auth_dict:
+        return
+
+    auth_dict = json.loads(auth_dict)
+    verification = auth_dict.pop('hash')
+
+    data_check_str = '\n'.join(
```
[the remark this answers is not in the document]
ajdavids: Will do.

web/app/notifications.py
```diff
@@ -117,10 +126,21 @@ def send_failure_alert_pushbullet(printer, is_warning, print_paused):
...
+def send_failure_alert_telegram(printer, is_warning, print_paused):
+    if not printer.user.telegram_eligible():
+        return
+
+    chat_id = printer.user.telegram_chat_id
+    notification_text = notification_elements(printer, is_warning, print_paused)
+
+    photo = get_photo(printer)
+
+    telegram_bot.send_notification(chat_id, notification_text, photo, printer_id)
```
kennethjiang: I got `NameError: name 'telegram_bot' is not defined` error on this line when I was testing it. Did it work for you?

Also shouldn't `printer_id` be `printer.id` instead?

web/app/telegram_bot.py
```diff
@@ -0,0 +1,147 @@
...
+    users = User.objects.filter(telegram_secret=secret)
+
+    if len(users) is not 1:
+        LOGGER.warning('Callback called with invalid secret.')
+        raise
+
+    return users[0]
+
+def get_printer(user, printer_id):
+    printer = Printer.objects.filter(user=user, id=printer_id)
```
kennethjiang: Should this be `printers` instead of `printer`?

web/app/telegram_bot.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if len(printers) is not 1:
+        LOGGER.warning('Callback called with invalid printer id.')
+        raise
```
kennethjiang: I'm pretty sure you can't just use `raise` without an exception object but I could be wrong.

web/app/forms.py
kennethjiang: I assume it is used when we send alert notification about a printer. If I'm right here, we already have printer at that moment and we can just use `secret_token` of that printer. But I could be wrong here since I'm still trying to figure out where `telegram_secret` is used and how it can secure the communication.

kennethjiang:
> How the telegram secret works overall is just as something temporary to let us verify callbacks. Instead of the DB, I can stick a random string in redis whenever a notification is sent and remove it whenever an action is taken. That would be more secure too, I think. I'll do that.

Can you send me a link to Telegram's document related to how this handshake works? Better yet, we probably want to insert that link as a comment in the code to help explain it.

ajdavids:
https://core.telegram.org/bots#deep-linking-example - I'm basing it on that, except that instead of using the secret as a payload to a telegram message, I'm using it as callback data on the inline keyboard.

kennethjiang:
> https://core.telegram.org/bots#deep-linking-example - I'm basing it on that, except that instead of using the secret as a payload to a telegram message, I'm using it as callback data on the inline keyboard.

Can we follow what telegram doc for this process? we can use redis in place of memcache.

ajdavids:
Yeah, that's the change I'm currently making.

ajdavids:
(see https://github.com/TheSpaghettiDetective/TheSpaghettiDetective/pull/87#issuecomment-522236177 )

web/app/telegram_bot.py
ajdavids: Good catch

web/app/notifications.py
ajdavids: Ah, I'd mostly tested it through sticking send_notification calls into the printers view for testing, since I haven't hooked my dev copy up to my printer yet. Right now going through and making the suggested changes.

ajdavids:
I made all  requested changes and tested more thoroughly. Thanks for the tip about `shell_plus`!

PR #112, opened 2019-11-04
Jetson support

RaymondHimle:
Adds dockerfiles for ARM64. Changes ml_api/Dockerfile to use correct darknet library based on detected system architecture. Separates requirements.txt into common requirements and architecture specific requirements. A manifest file for ml_api and web can make deployment to Jetson as simple as standard PC hardware.


ml_api/requirements.txt
```diff
@@ -1,9 +1,7 @@
 ipdb
-opencv-python>=3.4
```
kennethjiang: Any reason why we want to move opencv related module to a separate requirement file?

RaymondHimle: pip doesn't have opencv modules for ARM in their repo. The ubuntu-ports python-opencv worked as an alternative. I moved them so only x86 based machines try to download them with pip. I figured it was easier than finding an alternative pip package that supported x86 and ARM.

kennethjiang:
Thank you @RaymondHimle !

PR #113, opened 2019-11-05
Use Docker Compose Overide file for Jetson Settings

RaymondHimle:
Use docker-compose.override.yml for Nvidia Jetson specific configuration. Includes web to be rebuilt using the web dockerfile.base as thespaghettidetective/web does not have an ARM64 image.


kennethjiang:
Nice PR! Thanks @RaymondHimle

PR #118, opened 2019-11-20
Replace RawGit Gauge CDN with jsDelivr CDN

arcreigh:
RawGit is shutting down very soon,
jsDelivr will host the gauge javascript files.


arcreigh:
Test is checking out, gauge is showing up and working as intended.

PR #121, opened 2019-11-27
Fix E-Mail Notification link.

arcreigh:
This pull should fix the issue where alerts would be muted and TSD would no longer monitor a print if EU clicks on the resume link.

Line 55 initially had mute_alert set to true. Removing that flag from the URL here does fix the issue at least after my testing.


kennethjiang:
Thank you @arcreigh  for this awesome PR.

Issue #127, opened 2019-12-09
Email templates don't refer to private server


marcelrv:
The email templates in  `/web/app/templates/email/`
have hardcoded link to `https://app.thespaghettidetective.com/prints/` instead of using something like `site.build_full_url('/prints/')`

This means that the link will bring you to the wrong print

marcelrv:
Also seems like the new templates on private server complain about the low on detective hours

` You are running low on Detective Hours. Get more if you want The Detective to keep watching your prints.`

kennethjiang:
I can totally see how that happens! Need to think a bit more about how Detective Hours should be dealt with on private servers...

kennethjiang:
Already fixed.

PR #130, opened 2019-12-10
updated links in email

marcelrv:
addressing #127 

note. not sure of task.py is correct yet


kennethjiang:
I think it should work in `tasks.py` too. Any chance you can test that so I don't need to. :) Quite overwhelmed these days...

marcelrv:
sorry for the lack of response... 
Indeed the change on  task.py is not working correct yet.

it fails on site object even though the import is there. Dunno how to work around that

PR #144, opened 2019-12-28
Fix WS proxy

RaymondHimle:


kennethjiang:
Can one of the admins verify this patch?

PR #145, opened 2019-12-28
Request a new telegram connection every time.

ajdavids:
Request a new telegram connection every time instead of only when telegram_bot.py is first included. This avoids timeouts.


kennethjiang:
Can one of the admins verify this patch?

kennethjiang:
Can one of the admins verify this patch?

PR #159, opened 2020-01-26
Fixed bug in timelapse_video.overlay_detections

diveboatslave:
There's a small bug on the drawing of the detection overlays.


RaymondHimle:
As this is currently it will break compatibility with Jetson Nano. Any container used must be built for both ARM64 and AMD64

diveboatslave:
Sorry, I pushed 4 subsequent commits onto the main branch yesterday, and forgot this would pollute my PR. I'll move it onto a different branch.

PR #220, opened 2020-06-15
Extended example with image references

encetamasb:
And some eslint fixes.


web/frontend/src/print_shot_feedback/App.vue
```diff
@@ -14,7 +14,11 @@
       />
       <card v-if="print !== null && currentShot && print.access_consented_at">
         <div class="current-shot-container">
-          <img class="card-img-top" @click="next" :src="currentShot.image_url" />
+          <img
+            class="card-img-top"
+            @click="next"
+            :src="currentShot.image_url"
+          >
```
kennethjiang, line 21: It looks like your auto-format tool (I don't think you did it yourself) replaced `/>` with `>`, which makes it not well-formed.

encetamasb: vue/html-self-closing: Disallow self-closing on HTML void elements

encetamasb:
I've removed all non-related eslint changes.

PR #223, opened 2020-06-15
Switched to vue-sweetalert2

encetamasb:
Working toast and position.


encetamasb:
Please check @kennethjiang

PR #227, opened 2020-06-19
 Added bootstrap/bootswatch/bootstrap-vue to main.sass 

encetamasb:
Removed vendored version of bootstrap as well.


encetamasb:
Building time gets sensibly longer with this change.

Is there anything missing from usual look n feel? Are bootsrap-vue components still off (if yes which one)? @kennethjiang

kennethjiang:
It worked beautifully. Longer build time is not a big deal.

PR #228, opened 2020-06-19
Optimized moment.js locales out (smaller & faster builds)

encetamasb:


kennethjiang:
This does cut the bundle size significantly. 

The bundle size of `prints.js` is still over 1MB. I'm sure it contains the duplicate js libs with `print_shot_feedback.js`.

I think the ultimate solution of the bundle size (which I care a lot more than the build time) is to bundle all the libs into a separate output that is shared across all pages. I remember webpack can be configured to do so but I could be wrong. Do you want to take a stab at it next week?

I'm leaving this PR open for now as a thread for us to talk about the solution.

PR #233, opened 2020-06-23
Fullscreen print video playback in modal with gauge

encetamasb:


web/frontend/src/prints/PrintCard.vue
```diff
@@ -60,9 +60,16 @@
...
-      <div v-show="cardView == 'info'">
-        <video-box :videoUrl="print.video_url" :posterUrl="print.poster_url" />
-        <div class="card-body">
+      <div>
+        <video-box
+          :videoUrl="print.tagged_video_url"
+          :posterUrl="print.poster_url"
+          :fluid="true"
+          @timeupdate="onTimeUpdate"
+          @fullscreen="$emit('fullscreen', print.id)"
```
kennethjiang, line 69: Is this the only line that you changed to `PrintCard.vue`? When I pulled, there were changes and conflicts which made it quite difficult to tell if any bug was created in the process.

encetamasb: I recreated changes on your master, without real "rebase". Somehow git rebase could not create proper diffs for me so I went with the manual way. The unreadable change itself is rather simple:
<div> V X </div> <div> V Y</div> => <div> V <div> X </div> <div> Y </div> </div>

Issue #203, opened 2020-06-04
[FEATURE REQUEST] Indicator for Maximum spaghetti detected on a print


serenewaffles:
It would be nice if TSD indicated the maximum spaghetti detected on a print using the gauge already implemented. I imagine this happening similar to mechanical max reading gauges.

As can be seen in the below image, the black hand represents the current reading. When it moves higher, it catches the red hand and moves it as well. The red hand is not spring loaded, so it does not return. I imagine having a small triangle on the outer part of the TSD dial to indicate it as well. This could help users determine whether review is necessary for ongoing prints. Please see the second picture for an example.

![unnamed](https://user-images.githubusercontent.com/775693/83763572-dafa9c00-a646-11ea-889a-37dd693d0804.jpg)

![image](https://user-images.githubusercontent.com/775693/83763822-24e38200-a647-11ea-971d-0eabf9e158d8.png)

kennethjiang:
Really good suggestion.

What will be a good way to help new users understand the small triangle means the max value?

serenewaffles:
If we're concerned about how a new user would understand, I recommend either a help text popup (on hover) or gating it behind an option. Another option could be to have a quick popup for new users (or for all users after this is implemented) that has a simple picture and explanation.

kennethjiang:
I guess I don't know how to articulate the benefit of this new piece of info to users. In other words, I'm struggling to figure out how we can tell users something like "this 'max detection' means X. You can use it to do Y. Take Z action when this value is too high/low."

serenewaffles:
Having played more with Focused Feedback, I'm no longer certain this addition offers any benefit. Focused Feedback is already alerting users to frames that TSD is unsure of.

PR #240, opened 2020-06-28
Display max prediction value for gauges

encetamasb:
Simple solution for #203

### DIFF
--- a/web/frontend/src/common/Gauge.vue
+++ b/web/frontend/src/common/Gauge.vue
@@ -1,6 +1,7 @@
 <template>
   <div class="tsd-gauge-container">
     <span id="title" :style="{color: titleColor}">{{ titleText }}</span>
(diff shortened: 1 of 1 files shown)


kennethjiang:
Thank you for donating this PR. It's definitely functional but I'm still afraid that the benefit of this addition doesn't justify the potential confusion is causes to the users. A lot of times a better user experience is achieved when less info, not more, is shown. And I suspect this is the case.

My personal opinion is most users will be confused by what this "max" means (life time max? max of his printer? or max of this print so far?), why they would need this info, and what action they can take based on this info.

Let me close this PR for now. I'll check with our UX/UI advisor next time I meet him. If he has a good reason why we want to have this, I'll reopen and merge it.

One tiny thing that is not that important - sometimes the "max" can be great than 100, which would make the tag to go beyond its `<div>`. I'm pretty sure it can be easily fixed if we decide to bring back this PR and merge it. No action is needed now.

Thanks again @encetamasb for contributing this PR. Sorry I don't feel good at merging it - not because of your code, but because my overall consideration on the impact on the user experience.

PR #263, opened 2020-08-04
Update docker-compose.yml

gchedzhemov:
Remove extra spaces before commented out keys (after #) in order to reduce user confusion when setting keys commented out by default: https://discordapp.com/channels/614543405724205137/648759742881071124/740226474895736972


kennethjiang:
Good call! Thank you so much George!

PR #290, opened 2020-09-15
Made reason of "not watching warning" visible

encetamasb:


kennethjiang:
It worked! really elegant solution!

I do realize there were some confusing terms in the code existed before your PR and probably made it difficult for this PR. 
I'm thinking about refactoring them with this PR to set us up for a slightly more maintainable code.

* `watching` field in the Printer model. This is misleading as it represent a different meaning than the "watching" status. Maybe rename it to `enable_watching` in the model to make it more clear?
* `should_watch` virtual field in the Printer model. I think this is closer to what "watching" means to the users. Maybe rename it to `watching`? But I'm ok with leaving its name as is.

PR #325, opened 2020-12-02
Heater target notifications [jsonfield version]

encetamasb:
If you liked this concept better...


kennethjiang:
I guess I misunderstood your proposal on the json field. I thought it would be a field on the `Printer` object. I don't see the point of having it on the Tracker object. I'll close this one. Let's go with the other PR.

PR #342, opened 2020-12-31
Improved octoprint status and event processing

encetamasb:
Prints should be considered done only when 
(1) explicit event (done/failed) is received from plugin or 
(2) when estimated print time (+ preconfigured delta) is reached without any update from plugin or 
(3) after a specified absolute delta from start when there is no estimated print time. 

To store necessary state the following db related changes are proposed:

`Print.status_at` -  datetime of last update from plugin 
`Print.estimated_finished_at` - last known estimate 
`Print.gone_at` - datetime when background job marked "forgotten" print as failed 

`PrintEvent.DETECTED` - ongoing print detected without a `PrintStarted` event (somehow it was missed)
`PrintEvent.GONE` - print received no updates for a while and was marked as failed by background job

Notes:

(1) Plugin instances sharing the same auth token (what's an anomaly) and tracking prints in parallel may work now but proper detection of print failure in these cases might be impossible without other changes. Anyway, in non worst case scenarios situation is improved.
(2) status_at and estimated_finished_at are not updated automatically on every state updates to ease db load (see `PRINT_TOUCH_DELTA` and `FORCE_ESTIMATE_DELTA`)
(3) Printer.current_print can change without closing the current print;  `PrintStarted` event always updates it, "healthy" prints are kept as current, edge cases are controlled by `UNSET_DELTA` and `CURRENT_SWITCH_OVER_DELTA`.
(4) improved event notifications (webhooks, mobile..) are not part of this PR, but could come as a natural next step.


web/api/octoprint_messages.py
```diff
@@ -48,46 +54,263 @@ def settings_dict(octoprint_settings):
...
+    # touching print and updating estimate
+    delta: datetime.timedelta = (now - (print.status_at or print.updated_at))
+    if print.ended_at() is None and delta > datetime.timedelta(PRINT_TOUCH_SECS):
+        print.status_at = now
+
+        next_estimate: Optional[datetime.datetime] = extract_estimate(op_status)
+        if next_estimate is not None:
+            print.estimated_finished_at = next_estimate
+
+        print.save(update_fields=('updated_at', 'status_at', 'estimated_finished_at'))
```
kennethjiang, line 297: Any way to make sure when in the future we change other field in the process, we will remember to add that field to this `update_fields` set?

encetamasb: Something like this might help that. Interested?

+class UpdateWrapper:
+
+    def __init__(self, obj: models.Model, force: Optional[Tuple[str]]) -> None:
+        self._w_obj = obj
+        self._w_force = force or ()
+        self._w_changes: Dict[str, Any] = {}
+
+    def __setattr__(self, key: str, value: Any) -> None:
+        if key.startswith('_w_'):
+            self.__dict__[key] = value
+            return
+
+        self._w_changes[key] = value
+        setattr(self._w_obj, key, value)
+
+    def __getattr__(self, key: str) -> Any:
+        if key.startswith('_w_'):
+            return self.__dict__[key]
+        return getattr(self._w_obj, key)
+
+    def save(self, **kwargs) -> None:
+        keys = tuple(set(self._w_changes.keys()) | set(self._w_force))
+        self._w_obj.save(update_fields=keys, **kwargs)

kennethjiang: If the only reason to do it is to save the data transferred for the update SQL, I'd say we skip the `update_fields` param. The overhead for the update SQL should be small enough to be almost negligible. Don't feel it's worth the extra complexity and chances for error.

encetamasb: I propose this to avoid data races. save() overwrites the row fully in the table and as these functions might alter only low number of fields this is how I would do it properly.

encetamasb: An other rather basic option is to collect changes in a dict and apply them together. After playing a bit with the idea this updatewrapper still looks nice but might be missing some points, ruins type checks and adds some extra trouble when one needs to reference original object for setting related fields (thought I can simply wrap everything in the top function and forget it).

web/app/tasks.py
```diff
@@ -33,6 +37,54 @@
 
 LOGGER = logging.getLogger(__name__)
 
+# if estimated_finished_at + this is in the past then
+# its time to emit PrintEvent.GONE
+GONE_DELTA = timedelta(hours=4)
+
+# if unclosed print without estimate has an older
+# updated_at than this then its time to close it (last resort)
+ABS_GONE_DELTA = timedelta(days=7)
```
encetamasb, line 46: or less? ~max expected print time before commit goes live

PR #344, opened 2021-01-03
Update jetson_guide.md

LyricPants66133:
Updated the guide to be more of a guide. Also, link at the beginning of doc does not work. I did not remove it, but it should be updated.


kennethjiang:
This is awesome! Let me know when you think this PR is ready to merge @LyricPants66133

LyricPants66133:
I think that the changes are ready to merge. The only thing that still needs changes is the link at the begging of docs/jetson_guide.md.

Issue #90, opened 2019-08-19
Pushover for notifications


flippedcracker:
If possible, I'd like to see Pushover available for notifications.

https://pushover.net/

Nimmsis:
I agree with this. pushover have a easy api, the ability to set priority (can even bypass do not disturb) and is fast a reliable.

byurick:
+1 to this especially since Pushbullet no longer has an iOS app available.

vikwin:
The feature was added with merge #357. This issue can be closed.

kennethjiang:
Haha yes! Thank you! I'm terrible at github hygiene. :)

PR #357, opened 2021-01-25
Added pushover integration as per #90

vikwin:
Added basic integration of Pushover API.

### DIFF
--- a/web/app/forms.py
+++ b/web/app/forms.py
@@ -51,7 +51,8 @@ class UserPreferencesForm(ModelForm):
         fields = ['first_name', 'last_name', 'phone_country_code', 'phone_number', 'pushbullet_access_token',
                   'telegram_chat_id', 'notify_on_done', 'notify_on_canceled', 'account_notification_by_email',
                   'print_notification_by_email', 'print_notification_by_pushbullet', 'print_notification_by_telegram',
-                  'alert_by_sms', 'alert_by_email', 'discord_webhook', 'print_notification_by_discord']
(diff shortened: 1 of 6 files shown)


kennethjiang:
That looks awesome!

Can you give some brief instructions on how I can set it up and test it in my local dev env? @vikwin

vikwin:
@kennethjiang Of course, these are the steps you need to do:

1. Register an account on pushover.net
2. On your account page, add an application. Most information is optional, just give it a name like "TSD" for example. This will create an App token.
3. Add a device (Android, iOS or Desktop). There's a 7 day trial for each device, after that you need to pay a small one time fee if you want to continue using pushover on that device. It's only like 5€ or so.
4. On TSD user preferences, add the pushover app token  and the pushover user token. The user token can be found on the pushover account page.
5. Let TSD do it's job, you should receive push notifications on your device.

kennethjiang:
@vikwin Sorry for taking so long to merge.

I made some changes before merging. The most important change is to move app_token to env vars. I made that change based on a few reasons:

1. Ease of use. TSD users don't have to jump through the hoops to set up their own apps and get token.
2. Consistency with the design pattern in other notification channels
3. No obvious use case when a TSD server would need one app per user.

However, it may mess up the database migration in your server. To avoid that problem, please do these:

1. Do pull from this repo.
2. run `docker-compose web migrate app 0046` # this will erase app token and user token you already have in your db.
3. `git pull`
4. edit `docker-compose.yaml` to set `PUSHOVER_APP_TOKEN` 
5. `docker-compose restart`

Let me know if you run into any issues.

PR #378, opened 2021-02-14
Add 'write' access request when logging in with telegram.

ajdavids:


ajdavids:
Fixes [Issue 377](https://github.com/TheSpaghettiDetective/TheSpaghettiDetective/issues/377)

PR #394, opened 2021-03-02
Updated README to include new TSD plugin Interface

ryryog25:
Updated to include new description of the plugin and setup process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ After you have closed the wizard or if you already had the TSD plguin installed:
 
 4. Click "Save". OctoPrint isn't necessarily smarter than Django after all.
 
-![Change Server Address](https://github.com/ryryog25/TSD-fork-resources/blob/main/Screen%20Shot%202021-03-02%20at%209.20.58%20AM.png?raw=true | width=250)
+<img src="https://github.com/ryryog25/TSD-fork-resources/blob/main/Screen%20Shot%202021-03-02%20at%209.20.58%20AM.png?raw=true" width="250">
 
 5. Now, after going back to The Spaghetti Detective plugin settings, click "Troubleshooting"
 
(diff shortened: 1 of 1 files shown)


ryryog25:
The only thing left is that the image below the instructions should be moved and should also be updated, I can do that now...

ryryog25:
Ok, I got the images formatted correctly, had to do some trial and error (I probably did it wrong, don't really know how to use GitHub.)

ryryog25:
I think that I have made all the changes that there should be, if there is anything else tell me!

README.md
```diff
@@ -109,25 +109,50 @@ That's it! You now have a fully-functional The Spaghetti Detective server that y
 
 # Configure The Spaghetti Detective OctoPrint Plugin to use your own server
 
-Before you can configure The Spaghetti Detective OctoPrint Plugin to use your own server, you need add a printer to The Spaghetti Detective server you just built and obtain the secret token for that. To do so:
+Before you can configure The Spaghetti Detective OctoPrint Plugin to use your own server, you need add a printer to The Spaghetti Detective server you just built and obtain the 6-digit Verification Code for that. To do so:
 
 1. Pointing your browser to `http://your_server_ip:3334`.
 
-1. Log in as a user (you can just login with `root@example.com` but it's more secure to use a non-admin user).
+2. Log in as a user (you can just login with `root@example.com` but it's more secure to use a non-admin user).
```
kennethjiang, line 116: Using `1.` is markdown's way to say these are numbered bullets. They will be displayed as the correct sequence numbers. The advantage of not hard-coding the sequence number is we don't have to change all the bullets if we decide to insert a line.

README.md
```diff
@@ -109,25 +109,50 @@ That's it! You now have a fully-functional The Spaghetti Detective server that y
...
+1. Open the OctoPrint Settings and navigate to "Access Anywhere - The Spaghetti Detective"
+
+2. Click "Settings"
+
+3. Find and change the Server Address to `http://your_server_ip:3334/` (use https:// if you have it, if you aren't sure, just use http://) you MUST include the "http://" and trailing "/"
 
 4. Click "Save". OctoPrint isn't necessarily smarter than Django after all.
 
-![Site configuration](https://raw.githubusercontent.com/TheSpaghettiDetective/TheSpaghettiDetective/master/docs/plugin_config.png)
+<img src="https://github.com/ryryog25/TSD-fork-resources/blob/main/Screen%20Shot%202021-03-02%20at%209.20.58%20AM.png?raw=true" width="500">
```
kennethjiang, line 136: Can you add the images as part of your PR? It's usually a good idea to include everything needed to be in the same repo to make it self-contained.

ryryog25: Sure!

README.md
ryryog25: Ok, so I should make all of them one?

ryryog25:
I can no longer edit the file, going to redo with just one pull request for everything.

PR #397, opened 2021-03-02
Changed README to have the new TSD plugin setup/interface

ryryog25:
I needed to create a new pull request because I could not find how to add files to a PR. I also lost the ability to edit the README.md in the pull request, as it would just go to a 404 error.


ryryog25:
Also, thanks for catching my oversight with the Advanced server configuration drop down, didn't even know that was there!

PR #398, opened 2021-03-02
Update Code

LyricPants66133:
I updated the code run to install the server as to work with a change I made to the Install Script.


kennethjiang:
Thanks again @LyricPants66133

PR #400, opened 2021-03-03
Update README.md and added files to docs to include steps to Re-run the setup wizard

ryryog25:
Just added two files to docs, a md file with instructions, a screenshot that is in the md, and updated the readme so that there is a link to the doc which would help users who have had the TSD plugin installed before with app.thespaghettidetective.com or something similar and do not have the setup wizard upon restart.


ryryog25:
Also, I can fix that part that I accidentally changed in the README that added a space, lol, just to make it cleaner...

PR #439, opened 2021-04-21
Helpers to get pending message counts of channel layers

encetamasb:


encetamasb:
Usage:

    In [1]: import lib.channels
    
    In [2]: lib.channels.get_all_channel_layer_msg_count()
    Out[2]: {'default': 688, 'octoprint': 0}

PR #443, opened 2021-04-25
Touch group membersip only every N secs

encetamasb:


kennethjiang:
Great PR. I replicated your code in this commit https://github.com/TheSpaghettiDetective/TheSpaghettiDetective/commit/bbabc1c423c611c931f37d4d9cc2642341727f44 instead since it is easier than merging your PR.

PR #448, opened 2021-05-04
Adding a new setting to include snapshots into email

sylvaneau:
I'm running TSD self hosted behind a proxy that add a layer of authentication.

The actual notifications system detects my server as a public one (because it is exposed by a public url) and then add a link to the snapshot into the email. But because of the requested authentication the link into the email is broken.

This new setting allows to force snapshots to be attached to the notification email. It does not change the default behavior but add extra flexibility for self hosted environements.

I hope you'll find this feature interesting, feel free to change any name if you think there aren't clear enough.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ x-web-defaults: &web-defaults
     EMAIL_HOST_PASSWORD:    # -> your email account password
     EMAIL_PORT: 587  # DO NOT surround it with quotes. Otherwise email won't be sent!
     EMAIL_USE_TLS: 'True'
+    EMAIL_INCLUDE_SNAPSHOTS: 'False'
     DEFAULT_FROM_EMAIL: 'changeme@example.com'
     DEBUG: 'False'    # Don't set DEBUG to True, otherwise the static files will be cached in browser until hard-refresh
     SITE_USES_HTTPS: 'False' # set it to 'True' if https is set up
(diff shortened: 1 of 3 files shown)


kennethjiang:
This PR makes a lot of sense. Thank you @sylvaneau

sylvaneau:
Hi,

Hmmm there is something wrong... I've spent the whole day trying to make it works but it doesn't... The received mail do not embed the snapshot, it seems that the new environment variable is ignored...

I've attached a debbuger to the python app inside the container, and set a breakpoint in send_email in notifications.py. When I used the django backend to send a test email, the breakpoint is trigered... but when the mail is sent at the end of a print job my breakpoint is not trigered anymore...

Do you I miss something ? I'm wondering if there is an issue with the "tasks" service because i can't figure out how the mail is sent when a print is done.

sylvaneau:
Got it !

We need to make a small change to this feature. the `img_url` passed to `send_email()` is actually built up from the server url stored in django db. So when the method tries to download the snapshot it uses this url, and in my case this url is portected by my reverse proxy (the famous authentication layer) so it fails to download the image.

I will try to use the `INTERNAL_MEDIA_HOST` environment value to build an internal url to download the image.

What do you think ? (Feel free to PM me in discord if i'm not clear enough)

kennethjiang:
INTERNAL_MEDIA_HOST was designed to have the private IP address that can only be accessed by other containers in the same cluster, instead of the end user.

The django "site" is the address that "the end user" can use, including the browser and the plugin.

How did you configure the "Server Address" in TSD plugin?

sylvaneau:
Another option, much more easier... is to embed snapshot as link into the email instead of as attached images. For now i temporary modify the mail templates to include a link for `img_url` instead of a `img` tag

PR #453, opened 2021-05-13
Keep presence state in redis instead of db

encetamasb:


kennethjiang:
Close it for now. We can revisit it after migration to GCP

PR #484, opened 2021-06-21
Add expected to finish at time

B-0-B-B-Y:
Personally have found that I like planning my printing around what time the current print will finish, and sure, you can calculate this yourself by looking at how many hours and minutes are expected for this print to take in the total print time section, but imo it's nicer to just have this calculated directly for you, so this PR adds a calculated end time to the total section, formatting it like so:

```total print time in hours and minutes | expected end time```

Example:

``` 4h56m | 16:34 ```


kennethjiang:
Thanks a lot for the PR. Can you attach a screenshot for how it looks like on a mobile phone (375x712)?

B-0-B-B-Y:
> 
> 
> Thanks a lot for the PR. Can you attach a screenshot for how it looks on a mobile phone (375x712)?

No worries, and sure!
Here it is, as rendered by chrome dev tools in responsive mode set to 375x712:

image

kennethjiang:
> > Thanks a lot for the PR. Can you attach a screenshot for how it looks on a mobile phone (375x712)?
> 
> No worries, and sure!
> Here it is, as rendered by chrome dev tools in responsive mode set to 375x712:
> 
> image

That's what I'm afraid of. Since most users are on their phones, we need a design that looks good on phones first...

B-0-B-B-Y:
> > > Thanks a lot for the PR. Can you attach a screenshot for how it looks on a mobile phone (375x712)?
> > 
> > No worries, and sure!
> > Here it is, as rendered by chrome dev tools in responsive mode set to 375x712:
> > 
> > image
> 
> That's what I'm afraid of. Since most users are on their phones, we need a design that looks good on phones first...

Ah okay, no worries, I can redesign it, not an issue 😄. For mobile, would you be happy to have the total time be clickable, so when you click on it, it would toggle between total hours and expected finish time? And I can add a small icon around the time or some kind of indicator to tell users that they can click on that area to switch between those 2 views?

kennethjiang:
@B-0-B-B-Y What you said sounds like a really elegant design! Can you make the first field to switch between remaining and total time by clicking, and the 2nd field to be "estimated"? It is more consistent with the mobile app design too.

kennethjiang:
Closing because of inactivity.

PR #485, opened 2021-06-22
Moved redis related discovery functions to cache module and added guard for blank public ip

encetamasb:


web/api/viewsets.py
```diff
@@ -435,7 +436,7 @@ class PrinterDiscoveryViewSet(viewsets.ViewSet):
     def query(self, request):
         client_ip, is_routable = get_client_ip(request)
```
encetamasb, line 437: This may result None, without an extra check this would be undetected

PR #518, opened 2021-09-29
Added phone_country_short and neccessary data for frontend

encetamasb:
Added country short codes and names to support work on an
enhanced phone number input.

Frontend should migrate and use TWILIO_COUNTRIES solely.


kennethjiang:
@encetamasb Dmitry and I decided to abandon the task as the package is buggy, and the extra complexity is not worth it. Closing your PR now.

PR #521, opened 2021-10-26
Tunnel v2

kennethjiang:


web/lib/cache.py
```diff
@@ -298,3 +298,14 @@ def disco_pop_raw_device_messages(
             tordset_key, min='-inf', max=cur_time - expiration_secs)
         conn.zpopmin(tordset_key, message_count)
         return [raw_msg for (raw_msg, _) in conn.execute()[1]]
+
+
+def printertunnel_transferkey_set(transferkey, session_key):
```
kennethjiang: Can this be deleted since we are no longer transferring keys?

web/config/settings.py
```diff
@@ -310,6 +313,16 @@
 SLACK_CLIENT_ID = None
 
 OCTOPRINT_TUNNEL_CAP = int(os.environ.get('OCTOPRINT_TUNNEL_CAP', '1099511627776'))  # 1TB by default
+OCTOPRINT_TUNNEL_SUBDOMAIN_FMT = '{subdomain_code}.tunnels.{site.domain}'
+OCTOPRINT_TUNNEL_SUBDOMAIN_RE = re.compile(r'^(\w+)\.tunnels.*$')
+USE_OCTOPRINT_TUNNEL_V2 = (
```
kennethjiang, line 318: What will be the time when USE_OCTOPRINT_TUNNEL_V2 should be set to False?

web/config/settings.py
```diff
@@ -310,6 +313,16 @@
 SLACK_CLIENT_ID = None
 
 OCTOPRINT_TUNNEL_CAP = int(os.environ.get('OCTOPRINT_TUNNEL_CAP', '1099511627776'))  # 1TB by default
+OCTOPRINT_TUNNEL_SUBDOMAIN_FMT = '{subdomain_code}.tunnels.{site.domain}'
```
kennethjiang, line 316: This is no longer used, right?

web/app/middleware.py
```diff
@@ -0,0 +1,34 @@
+from django.conf import settings
```
kennethjiang, line 1: Can we make TunnelV2 a separate app, and deploy it to a different container? That will eliminate the need for this middleware, or other places where we use if/else to switch between tunnel v1/v2.

web/lib/tunnelv2.py
```diff
@@ -0,0 +1,176 @@
...
+                # req is not using basic auth, no 401 error
+                raise django.http.Http404
+
+        # no basic auth, no session - force basic auth
+        # TODO: when we have proper integrations we might
+        # be able to convert this to a 404
+        raise TunnelAuthenticationError('missing credentials', realm=realm)
+
+    @classmethod
+    def is_tunnel_request(cls, s_or_r: ScopeOrRequest) -> bool:
```
kennethjiang, line 168: If we make TunnelV2 a separate app and a separate container, this is not necessary either.

kennethjiang (commented):
I am now leaning toward making TunnelV2 its own app/container. I think it's a much cleaner solution, and will make it easier if we want to make TunnelV2 a non-django app in the future.

web/config/settings.py
```diff
@@ -310,6 +313,16 @@
...
+OCTOPRINT_TUNNEL_SUBDOMAIN_RE = re.compile(r'^(\w+)\.tunnels.*$')
+USE_OCTOPRINT_TUNNEL_V2 = (
+    os.environ.get('USE_OCTOPRINT_TUNNEL_V2', 'False') == 'True'
+)
+
+if os.environ.get('OCTOPRINT_TUNNEL_PORT_RANGE'):
```
kennethjiang, line 322: Let's hard-code the range to something that's unlikely to conflict, such as 53377-53477. And hardcode this port mapping in docker-compose.yaml too, so that the private server can run tunneling without changing any files.

encetamasb: Let's hardcode it into the docker compose setup, then its possible to override that hardcode using env vars. Over 32k it's not very "safe", those are assigned dynamically, as far as I know. When a required port is not free, container does not start.

web/lib/tunnelv2.py
encetamasb: Let's do this later

web/config/settings.py
encetamasb: Sorry, I cannot see what's the problem. Please fix it if you think there is one.

web/config/settings.py
encetamasb: Probably we can remove this.

PR #522, opened 2021-11-03
Added printer_id checks

encetamasb:
Let's be defensive and allow only messages with the expected printer_id.


kennethjiang:
I think we have decided not to do printer_id check since you figured out the root cause of the security issue.

PR #568, opened 2022-02-01
Retire tunnel v1 support

encetamasb:


kennethjiang:
Looks good. I'd remove `v2` from the the code as there will be no v1 anymore. But not a big deal if it's too much work. It's your call if you want to do that.

After that, please merge the PR and test it on staging.

PR #576, opened 2022-02-22
Update Upgrade instructions

LyricPants66133:
Upgrade instructions were Linux only, so I updated them for any OS.


kennethjiang:
Thanks @LyricPants66133 !

PR #578, opened 2022-02-26
Simplify templates system

dipuzyrev:


web/app/templates/layout.html
```diff
@@ -59,83 +59,79 @@
 
   {% block body_top_ext %}{% endblock body_top_ext %}
   {% include "snippets/svg_sprite.html" %}
+
   {% block navbar %}
-  {% with view_name=request.resolver_match.view_name %}
```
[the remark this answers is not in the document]
kennethjiang: Nvm. I found a way to ignore whitespace in github. So we are fine here.

web/app/templates/account/login.html
```diff
@@ -8,102 +8,103 @@
 
 {% block content %}
 {% get_providers as socialaccount_providers %}
+<div class="container">
```
kennethjiang, line 11: This <div> can be moved into `account/base.html` to further DRY the code. But it's not a big deal.

kennethjiang (approved):

web/app/templates/account/login.html
dipuzyrev: I didn't find a way to do it without wrapper_container

web/app/templates/vue_layout.html
```diff
@@ -4,15 +4,10 @@
 {% block navbar %}
 {% endblock navbar %}
 
-{% block content_wrapper %}
-  {% include 'snippets/messages.html' %}
```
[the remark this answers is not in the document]
dipuzyrev: Yes, we still have it, it's in layout.html and included for both vue and non-vue pages

web/app/templates/account/login.html
kennethjiang: This is an example of how you can implement it.
Screen Shot 2022-02-26 at 1 25 57 PM

 But now I'm not convinced myself that this is a better code. Just wanted to show you a way that can be done. Please merge without making such change.

PR #588, opened 2022-03-06
HTTPS should be used instead of SSH for git clone in the guides

tautomer:
Not all users have their public keys set up on github.

`git clone git@github.com` will fail for many users, especially the ones who don't really use github.

### DIFF
--- a/docs/jetson_guide.md
+++ b/docs/jetson_guide.md
@@ -29,7 +29,7 @@ should also work but we haven't tried them.
 1. Get the code:
 
 ```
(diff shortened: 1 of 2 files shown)


LyricPants66133:
This looks like a good pr. Only thing is that the readme.md should probably be included.

kennethjiang:
Good PR. Thanks @tautomer 

@LyricPants66133 Good catch about the readme. I'll take a quick look to see if more changes are needed there.

PR #600, opened 2022-03-22
Begin making changes to add a callback for the delete button.

jreed1701:
Call back added should trigger a web socket connection back to
octoprint TSD plugin to call delete function.

Work In Progress. Issues to figure out:
*  When clicking the delete button, need to figure out how to get a
   web connection to each printer in GCodePage.vue. Right now,
   the connection is null because I don't have a printer id.

Questions:
*  When deleting GCode, should it be deleted on all printers?
   Or, a specfic printer? Can't assume only uploaded to one.

Please Note: Started work off of "release" branch because "master" was not working at the time. At some point, will need to change the base of this branch over to master.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,7 @@
 version: '2.4'
 
-x-web-defaults: &web-defaults
+x-web-defaults:
+  &web-defaults
   restart: unless-stopped
   build:
     context: web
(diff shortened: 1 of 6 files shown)


jreed1701:
My current goal is to click the delete button and see that I triggered the delete function on the octoprint plugin via a log file.  Right now, I can click the delete button and get the callback in GCodePage.vue to fire but I'm having trouble getting a web socket to the octoprint plugin going.  

Here's a link to the code snippet I refer to:
https://github.com/TheSpaghettiDetective/TheSpaghettiDetective/blob/261507f6dc6a0e24ad286e384890b9785f268c0b/web/frontend/src/printers/GCodesPage.vue#L338-L411

PR #604, opened 2022-03-28
Reorganized notification handlers into plugins

encetamasb:


web/app/models.py
```diff
@@ -808,3 +808,60 @@ def is_octoprint_connected(self):
...
+
+    notify_on_heater_status = models.BooleanField(blank=True, default=False)
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        unique_together = ('user', 'type')
+
+    def phone_country_code(self):
```
kennethjiang, line 857: This looks quite hacky. 3 options:

1. make SMS a special case and have it own data model.
2. move this code to the sms channel impl itself.
3. make 2 config fields: `config1` and `config2`. This may be a good option as there could be other channels that will need 2 config fields (such as id + secret)

web/api/octoprint_messages.py
```diff
@@ -40,6 +43,22 @@ def process_octoprint_status(printer: Printer, status: Dict) -> None:
 
     channels.send_status_to_web(printer.id)
 
+    op_event = status.get('octoprint_event', {})
```
kennethjiang, line 46: Shouldn't we put this block of code in https://github.com/TheSpaghettiDetective/TheSpaghettiDetective/blob/master/web/app/models.py#L575 ? I suspect it's currently send duplicate messages.

web/app/models.py
```diff
@@ -274,25 +274,50 @@ def update_current_print(self, filename, current_print_ts):
...
+        assert print
+
         if print.cancelled_at is None:
             print.finished_at = timezone.now()
             print.save()
 
         PrintEvent.create(print, PrintEvent.ENDED)
         self.send_should_watch_status()
 
+        event_name = 'PrintDone'
```
kennethjiang, line 298: Isn't this duplicate with https://github.com/TheSpaghettiDetective/TheSpaghettiDetective/pull/604/files#diff-f821bd2c29b6c78c53566d64c0e903b4875b3f53e6b3243da0ea75d0fc4bb2e4R46 ? I'm having a hard time to follow when these notifications are sent as it seems we have at least 3 places to send them.

web/app/models.py
```diff
@@ -274,25 +274,50 @@ def update_current_print(self, filename, current_print_ts):
...
 
-    def unset_current_print(self):
+    def unset_current_print(self, gone=False, failed=False, canceled=False):
         print = self.current_print
         self.current_print = None
         self.save()
 
         self.printerprediction.reset_for_new_print()
 
+        assert print
```
kennethjiang, line 289: What's the purpose of having `assert` in non-testing code?

kennethjiang (commented):
- Is Email missing?

web/api/octoprint_messages.py
encetamasb: It does not send duplicate messages. If you see duplicates, then v1 is still enabled for you. That's on purpose, removing v1 is one of the very last step. The two can coexist and can be tested on prod, as far as you don't deploy v2 frontend.

I did not attempt to rework how events and status updates are handled. I did that twice in the past and it was never merged, I guess because it's complicated. I identified 3 points to hook v2 into. This is one of those.

web/api/octoprint_messages.py
```diff
@@ -40,6 +43,22 @@ def process_octoprint_status(printer: Printer, status: Dict) -> None:
...
+    if op_event:
+        event_name = op_event.get('event_type', '')
+        event_data = op_event.get('data') or {}
+
+        if event_name not in ('PrintStarted', 'PrintFailed', 'PrintDone', 'PrintCancelled'):
+            # events above are handled when PrintEvent is created
+
+            queue_send_printer_notifications_task(
```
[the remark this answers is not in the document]
encetamasb: As soon as we remove v1, plugin event handling won't happen there (except email etc). If plugins need any timelapse generation related event, that should be done separately as a separate event.

The current end task (what generates timelapse etc) remains configurable.

web/api/octoprint_views.py
```diff
@@ -43,6 +44,17 @@
 ALERT_COOLDOWN_SECONDS = 120
 
 
+def send_failure_alert(printer: Printer, is_warning: bool, print_paused: bool) -> None:
+    send_failure_alert_v1(printer, is_warning=True, print_paused=False)
```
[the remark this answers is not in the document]
encetamasb: As far as v1 is enabled, yes. Disabling it is one of the last thing. While testing v2 you can disable it easily.

PR #618, opened 2022-04-10
update readme with warning

LyricPants66133:
this warning will serve to remind users that with our rapidly changing repo, they should always cross-check with the commands located here. This will help solve some issues users still following old guides are having.


kennethjiang:
Thank you so much @LyricPants66133 . And sorry for the delay!

PR #655, opened 2022-06-12
Update telegram guide

mksdev:
Telegram bot can be registered without need of https domain or using looptools. I have been using this method in my local setup for a few months and someone might find it useful too.

### DIFF
--- a/docs/telegram_guide.md
+++ b/docs/telegram_guide.md
@@ -1,5 +1,6 @@
 # Telegram Setup Guide
 
+## Link using Telegram Login Widget
 My Base: Fresh Install Ubuntu 20.04.
 
 You need to open a terminal with your device. I am doing this via SSH (I'm using Putty on windows).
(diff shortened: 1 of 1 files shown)


kennethjiang:
Thank you so much for this new method. It's obviously a better method. Can you update the guide to just using your method? I don't see the point of keeping the "login widget method".

mksdev:
Nope, this configures the chat that you want the bot (which is set with env variable in docker) to send messages. So one bot can send messages to multiple chat rooms which are defined by chat id that you get with login widget or using my method.

There is only one bot for obico instance, and each user gets it's own chat with him.

kennethjiang:
> Nope, this configures the chat that you want the bot (which is set with env variable in docker) to send messages. So one bot can send messages to multiple chat rooms which are defined by chat id that you get with login widget or using my method.
> 
> There is only one bot for obico instance, and each user gets it's own chat with him.

Cool. Go for it!

kennethjiang:
@mksdev I'll close this PR as we have consolidated `telegram_guide.md` with all other docs in the `website` folder. If you are still interested in helping with this guide, please open another PR against the moved file.

PR #660, opened 2022-06-18
Refactor streaming

dipuzyrev:


frontend/src/components/StreamingBox.vue
```diff
@@ -90,14 +100,20 @@
...
 import Janus from '@src/lib/janus'
-import ViewingThrottle from '@src/lib/viewing_throttle'
+import { StreamThrottle, STREAM_SESSION, STREAM_EVENT } from '@src/lib/stream-throttle'
+import { isLocalStorageSupported } from '@static/js/utils'
+
+const STREAM_STATUS = {
+  Unavailable: 'Unavailable',
+  Available: 'Available',
+  Started: 'Started',
+}
```
dipuzyrev, line 112: Main refactoring idea is here. Previously we had `isVideoAvailable` and `isVideoVisible` variables, defining the state of the video. But they doesn't describe it clearly (in my opinion).

Now we have 3 main states of the video:
1. Unavailable
2. Available – webrtc steam is ready and can be attached to the <video /> tag
3. Started – stream is attached to the <video /> and will be played right after downloading enough part

At the 3rd step video could be in loading status (`isVideoLoading`) or in buffering (`isTrackMuted`)

frontend/src/components/StreamingBox.vue
```diff
@@ -147,28 +162,26 @@ export default {
 
   data() {
     return {
-      stickyStreamingSrc: null,
-      isVideoAvailable: false,
-      isVideoVisible: false,
-      remainingSecondsCurrentVideoCycle: 30,
-      remainingSecondsUntilNextCycle: -1,
-      currentBitrate: null,
+      STREAM_STATUS,
```
dipuzyrev, line 165: Include STREAM_STATUS so we can use it inside vue template

frontend/src/components/StreamingBox.vue
```diff
@@ -147,28 +162,26 @@ export default {
...
+      isTrackMuted: false,
+      isVideoLoading: false,
+
       slowLinkLoss: 0,
-      slowLinkShowing: false, // show on mousenter
-      slowLinkHiding: false, // hide on moseleave
-      trackMuted: false,
-      videoLoading: false,
-      printerStockImgSrc: '#svg-3d-printer'
+      slowLinkLossLimit: 50,
```
dipuzyrev, line 171: Define limit here in single place. Also we may want to use it in other places in the future

frontend/src/components/StreamingBox.vue
```diff
@@ -147,28 +162,26 @@ export default {
...
+      isSlowLinkShowing: false,  // show on mousenter
+      isSlowLinkHiding: false,  // hide on moseleave
+      currentBitrate: null,
+
+      remainingSeconds: StreamThrottle.defaultRemainingSeconds(),
+      activeSessionRemainingSeconds: StreamThrottle.calculateSessionsSeconds()[0],
+      frozenSessionRemainingSeconds: StreamThrottle.calculateSessionsSeconds()[1],
+      basicStreamTimerId: null,
+      currentSession: STREAM_SESSION.Idle,
```
dipuzyrev, line 180: Tried to encapsulate basic streaming timer logic with all the math as much as possible

frontend/src/components/StreamingBox.vue
```diff
@@ -147,28 +162,26 @@ export default {
...
     }
   },
 
   computed: {
-    taggedImgAvailable() {
-      return this.taggedSrc !== this.printerStockImgSrc
-    },
```
dipuzyrev, line 168: Unrelevant now

frontend/src/components/StreamingBox.vue
```diff
@@ -199,78 +212,89 @@ export default {
...
     },
     isBasicStreamingReadyToPlay() {
-      return !this.isVideoVisible && (this.remainingSecondsCurrentVideoCycle == 30 || this.remainingSecondsUntilNextCycle == 0)
+      return !this.autoplay && this.streamStatus === STREAM_STATUS.Available && !this.isTrackMuted && !this.isVideoLoading &&
+        this.currentSession === STREAM_SESSION.Idle
     },
     isBasicStreamingFrozen() {
-      return this.remainingSecondsUntilNextCycle > 0 && !this.isVideoVisible
+      return !this.autoplay && this.streamStatus === STREAM_STATUS.Available && this.currentSession === STREAM_SESSION.Frozen
     },
```
dipuzyrev, line 229: Just get rid of math as we do it in separate file

frontend/src/components/StreamingBox.vue
```diff
@@ -199,78 +212,89 @@ export default {
...
+      this.frozenSessionRemainingSeconds = frozen;
+
+      const whatHappened = StreamThrottle.whatHappened(newSeconds);
+      if (whatHappened === STREAM_EVENT.FrozenSessionStarted) {
+        this.webrtc.stopStream();
+      } else if (whatHappened === STREAM_EVENT.CycleFinished) {
+        clearInterval(this.basicStreamTimerId);
+        this.remainingSeconds = StreamThrottle.defaultRemainingSeconds();
+        this.streamThrottle.removeFromStorage();
       }
```
dipuzyrev, line 246: Probably not the best way to handle basic streaming timer but it makes mobile app and web app share exact same logic

frontend/src/components/StreamingBox.vue
```diff
@@ -15,70 +17,78 @@
...
       class="centered-element p-0"
       :disabled="isBasicStreamingFrozen"
     >
       <i class="fas fa-play ml-1" v-if="isBasicStreamingReadyToPlay"></i>
-      <span class="medium text-bold" v-if="isBasicStreamingFrozen">{{remainingSecondsUntilNextCycle}}s</span>
+      <span class="medium text-bold" v-if="isBasicStreamingFrozen">{{frozenSessionRemainingSeconds}}s</span>
     </b-button>
-    <b-spinner v-if="trackMuted || videoLoading" class="centered-element" label="Buffering..."></b-spinner>
 
-    <div v-if="isVideoAvailable">
```
dipuzyrev, line 37: Get rid of parent block

frontend/src/lib/stream-throttle.js
```diff
@@ -0,0 +1,105 @@
+const ACTIVE_SESSION_LENGTH = 30;
+const FROZEN_SESSION_LENGTH = 30;
```
dipuzyrev, line 2: Defining it in single place give us possibility to easily change duration of active or frozen part of basic streaming. Also we may want make active session longer or shorter than frozen session (this is because they are defined such way). Also, as mobile app have exact same file, it will be easily changed for both apps.

frontend/src/lib/stream-throttle.js
```diff
@@ -0,0 +1,105 @@
...
+
+  restoreRemainingSeconds() {
+    if (!this.storageProvider) {
+      return STREAM_CYCLE_LENGTH;
+    }
+
+    return this._calcRemainingSeconds(this.storageProvider.getItem(this.storageItemId));
+  }
+
+  async restoreRemainingSecondsAsync() {
```
dipuzyrev, line 34: for RN AsyncStorage

PR #663, opened 2022-07-05
Update README with Bonjour/Avahi Instructions

LyricPants66133:
Updated the readme to include instructions on how to set up Bonjour/Avahi on all systems for easier server connection. Will write up a script intended for Jetson that will set up avahi automatically, and will prompt user if they want to set their hostname to `obico`


kennethjiang:
Good PR. The only suggestion is to add this part to `website/docs/server-guides/configure.md`, or a separate file in the same folder instead.

I have intended to keep README to the bare minimum of what it takes to set up a server. All other advanced or alternatives documents go to `website/docs/server-guides/`, which will be pushed to "https://obico.io/docs/server-guides/".

PR #664, opened 2022-07-07
Avahi script for Jetson & Doc improvements

LyricPants66133:
Added a script that installs and sets up Avahi on the Jetson at server setup. User can opt out.

Also improved configuration docs for linux as avahi needs some setup to be good.

Fixed broken link in readme to avahi setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Upon fresh installation, the Obico Server can only work on the localhost. You wi
 
 ## Obtain server's IP address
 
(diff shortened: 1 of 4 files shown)


LyricPants66133:
Shell script files were giving me errors when set to CRLF, so I conveted them to LF to avoid any issues. These files are also supposed to be run on Linux, were LF is the standard.

kennethjiang:
Good job @LyricPants66133

PR #669, opened 2022-07-18
admin ip whitleisting is disabled only when the ip list is None

encetamasb:
None means whitelisting disabled, [] means admin is inaccessible (no ip-s
allowed).


kennethjiang:
I don't think it's necessary to differentiate between [] and None. I don't see a legitimate use case for enabling the ip whitelist but set it to an empty list.

Sorry I initially misunderstood this part.

PR #673, opened 2022-07-27
PrintDone handling fix (OctoKlipper)

encetamasb:
Job data is empty in status updates after PrintDone when using octoklipper.
Filename of job might be under data key in event metadata.


kennethjiang:
@encetamasb I fixed it with https://github.com/TheSpaghettiDetective/obico-server/commit/96a9a14688ece03f6803c7cb11b8f065bd1012ae

PR #682, opened 2022-08-24
Disallow multiple connections using the same token

encetamasb:


encetamasb:
@kennethjiang This is the core idea. Unideal part is the +1 layer group message per new connection. 

There is a custom error code (4321) hiding in the commit, at plugin side that condition can be handled. With current plugins the printers are going to keep reconnecting and kicking each other out every time. I would wait for at least 2 connect-disconnect cycles with 4321 error code, and then report that for the user (popup in octoprint).

kennethjiang:
@encetamasb It looks good. Can you make the change to the plugin side? I think in case of error code 4321, the plugin should just stop reconnecting, and show the error.

kennethjiang:
@encetamasb Pushed to master but github didn't close it automatically.

PR #683, opened 2022-08-24
Modify .env instead of compose file

zayigo:
To keep consistency with the [email setup](https://www.obico.io/docs/server-guides/configure/#email-smtp) I think it would be better to set all parameters in the `.env` file instead of telling the user to modify `docker-compose.yml` (arguably more sensitive to errors)


kennethjiang:
Thank you @zayigo !

PR #684, opened 2022-08-25
Fix typo

Rai-Oliveira:
Line 42 said "More likely than not, it's because your email configurations are set correctly.", which doesn't sound right, I assume it was a tiny mistake.


kennethjiang:
Thanks @Rai-Oliveira

PR #686, opened 2022-08-26
@report_error decorator for consumer message handlers

encetamasb:


backend/api/consumers.py
```diff
@@ -135,6 +157,12 @@ def get_printer(self):
         ).printer
 
     @newrelic.agent.background_task()
+    @close_on_error
+    def connect(self):
```
kennethjiang, line 161: Any reason why we redefine `connect` here?

backend/api/consumers.py
```diff
@@ -231,72 +260,55 @@ def disconnect(self, close_code):
             )
 
     @newrelic.agent.background_task()
+    @report_error
+    @close_on_error(exc_class=ObjectDoesNotExist, sentry=False)  # FIXME pls confirm that no need for sentry here
```
kennethjiang, line 264: Yes confirmed that `ObjectDoesNotExist` doesn't need to be logged in sentry

backend/api/consumers.py
encetamasb: Nothing else than consistency.

PR #692, opened 2022-09-02
Fixes for consumer error reporting

encetamasb:
Removed one connect method override added by mistake, this fixes shared printers.

Added additional @close_on_error rule.

### DIFF
--- a/backend/api/consumers.py
+++ b/backend/api/consumers.py
@@ -157,11 +157,6 @@ class SharedWebConsumer(WebConsumer):
         ).printer
 
     @newrelic.agent.background_task()
-    @close_on_error
-    def connect(self):
-        pass
-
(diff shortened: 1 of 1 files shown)


backend/api/consumers.py
```diff
@@ -201,6 +196,7 @@ def get_printer(self):
 
     @newrelic.agent.background_task()
     @close_on_error
+    @close_on_error(exc_class=Printer.DoesNotExist, sentry=False)
```
kennethjiang, line 199: Do you think `OctoprintTunnelWebConsumer` needs similar change? If I remember correctly, it also relies on the db query in "connect" to authenticate.

encetamasb: Yes. Adding.

PR #694, opened 2022-09-03
Added axis inversion feature for X/Y/Z

rahools:
As the title suggests, I have added an option to invert X/Y/Z motion inversion on the printer's setting page.

currently, the printer's setting page looks like this:

image


rahools:
It is not required. You achieve the same by configuring the appropriate settings in Octoprint.

PR #717, opened 2022-11-14
Found a letter lost a sea

cln-io:


kennethjiang:
Haha! Than you @cln-io

PR #723, opened 2022-11-23
Update ops.md

zanovis:
Added clarity for docker compose command


kennethjiang:
Thank you so much @zanovis

PR #736, opened 2022-12-29
Serve web container with daphne

gabe565:
Hello! I've been using Obico for a while and love it.

I was taking a look at the selfhosted stack and noticed the web container is being served with `python manage.py runserver`, which should not be used in production. According to the [Django docs](https://docs.djangoproject.com/en/4.1/ref/django-admin/#runserver):

> DO NOT USE THIS SERVER IN A PRODUCTION SETTING. It has not gone through security audits or performance tests. (And that’s how it’s gonna stay. We’re in the business of making web frameworks, not web servers, so improving this server to be able to handle a production environment is outside the scope of Django.)

Gunicorn won't work due to the websockets requirement, but Daphne has been working well for me.

Feel free to close this if there's a benefit to the built-in server that I missed.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,7 +59,7 @@ services:
       - "3334:3334"
     depends_on:
       - ml_api
-    command: sh -c 'python manage.py migrate && python manage.py collectstatic -v 2 --noinput && python manage.py runserver --nostatic --noreload 0.0.0.0:3334'
+    command: sh -c 'python manage.py migrate && python manage.py collectstatic -v 2 --noinput && daphne --bind 0.0.0.0 --port 3334 config.asgi:application'
 
   tasks:
     <<: *web-defaults


kennethjiang:
Sorry for the delay!

My only hesitation is daphne is only a ["reference server"](https://docs.djangoproject.com/en/4.1/howto/deployment/asgi/daphne/), which usually means it was created to demonstrate functions, with little consideration for performance or security. I believe this essentially puts it in the same bucket as the django dev server, but only used/exercised a lot less.

Therefore I have a feeling that daphne is even less prod-ready than django dev server. I haven't done much research on this so I could be wrong here. If you have found info that indicates otherwise, please let me know.

gabe565:
Hmm that's a good point. I haven't found anything explicitly stating Daphne is hardened and production-ready. I guess I was thinking that since the Django dev server explicitly says it _isn't_ production ready and the Daphne documentation doesn't say that, then Daphne may be more production-ready than the dev server. But you have a good point that the dev servers gets much more usage and is older, so may be more prod ready. That all-caps warning just makes me nervous...

encetamasb:
@gabe565 It uses daphne. 

See https://github.com/django/channels/blob/2.x/channels/management/commands/runserver.py#L6

In latest channels runserver command is no longer there though.

gabe565:
@encetamasb Good catch. Definitely better to leave this as is. Thank you both!

PR #737, opened 2022-12-31
Update telegram_guide.md

skruglov2023:
Fixed typo, added `apt` to installation commands, though could use a disclaimer than it can also be `apt-get` or something else


kennethjiang:
Thank you @skruglov2023

PR #746, opened 2023-01-23
Add Webhook Notification Plugin

lnjustin:
This PR is ready for review. It adds the 'webhook' notification plugin. The new webhook plugin has been tested for all types of notifications and found to be in working order. Thanks @encetamasb for all the help!


lnjustin:
I opted for a single Webhook endpoint, with the assumption that the user can just parse the JSON to determine the type of event. That suits my use cases just fine. But, if you think the predominant use case from your experience would require specifying dedicated webhook endpoints per event type, I can modify if preferred.

kennethjiang:
It looks pretty good. Can you open this against the "master" branch @lnjustin ? Sorry for letting you jump through another hoop!

lnjustin:
Sure, no prob.

PR #750, opened 2023-01-27
Adding another method to run obico-server using prebuilt images in docker-compose.yml

iluvatyr:
I added 

1) A modified docker-compose.yml named docker-compose-prebuilt.yml, which instead of building the images within the docker-compose up command refers to prebuilt images. Those images can be either built
- manually, using default docker build commands
- automated, using the script that is provided within this pull request (tested on Linux that supports bash). The script additionally gives the option to automatically run the obico-server after the images are built successfully.


kennethjiang:
What's the purpose of this PR?

PR #751, opened 2023-01-28
Clarify that MainsailOS x64 is not supported for webcam streaming

MallocArray:


kennethjiang:
Thank you @MallocArray . I do intend to make it work for the 64bit mainsailos since now it's released.

MallocArray:
> Thank you @MallocArray . I do intend to make it work for the 64bit mainsailos soon since now it's released.

Whew!  I was trying to take a look at it myself.  I see that you include a particular version of ffmpeg for x86 raspberry pi if the version detected is Debian 11, which would need to be updated to support x64, but also the version of janus in the repo is only x86. 
I made a Docker container to try to build my own janus x64 from source, and finally got it to compile, but when I tried to replace the current moonraker-obico version with it and use run_janis.sh I got an error I'm not qualified to troubleshoot any further:
`/home/pi/moonraker-obico/moonraker_obico/bin/janus/bin/janus: error while loading shared libraries: libconfig.so.9: wrong ELF class: ELFCLASS32`

So if you will look at it sometime, I'll just hang tight since I don't really know anything about Janus, and just keep on my MainsailOX x64 without streaming for now.

kennethjiang:
Thank you for helping me on this task! I certainly appreciate it as I'm quite overwhelmed right now with the new features (will come up soon!).

I originally committed the janus binary into git out of the habit of doing the same thing in OctoPrint, forgetting that the reason why I chose to do so for Octoprint was the installation of an octoprint plugin was inside a non-privileged process and hence it was difficult to install a system package during the process.

However, this is no longer true for Moonraker. I think the right thing to do in moonraker-obico is to install janus as a system package, unless we have a good reason why it won't work. I have started in this direction for ffmpeg: https://github.com/TheSpaghettiDetective/moonraker-obico/blob/master/moonraker_obico/webcam_stream.py#L63 it basically says we use "ffmpeg" in the system path unless it's debian 11 or later (I made an exception there because ffmpeg in these versions had a bug).

So I'd propose changing the installation script to install janus as a system package; and changing the code to use janus in the system package. Of course we want sufficient testing on various RPi OS versions to make sure janus system package does work.

MallocArray:
I wasn't able to find any system packages for janus that I could install in MainsailOS or a clean Debian container, but all for it if it exists. 

Also looks like several new releases of the same version of ffmpeg and newer major versions are out, so maybe those can be utilized in Debian 11 at this point as well

kennethjiang:
I am still running on an older version of mainsailOS but this is what I get:

```
pi@mainsail:~ $ apt-cache search janus
...
janus - general purpose WebRTC gateway
janus-dev - general purpose WebRTC gateway - development files
janus-doc - Open Source, general purpose, WebRTC gateway - documentation
janus-tools - Open Source, general purpose, WebRTC gateway - helper tools
libjs-janus - Open Source, general purpose, WebRTC gateway - JavaScript library
janus-demos - Open Source, general purpose, WebRTC gateway - demos
...
```

The `janus - general purpose WebRTC gateway` package should be the official janus package.

MallocArray:
No luck in MainsailOS x64
```
pi@mainsailos:~ $ apt-cache search janus
gnurobbo - logic game ported from ATARI XE/XL
gnurobbo-data - logic game ported from ATARI XE/XL - data files
python3-janus - thread-safe asyncio-aware queue for Python
librust-enum-to-u8-slice-derive-dev - Simple fork of enum_to_str_derive (by @DCjanus), convert enum to u8 slice ref - Rust source code
```

PR #752, opened 2023-01-30
Add Webhook Notification

lnjustin:
Adds the 'webhook' notification plugin. The new webhook plugin has been tested for all types of notifications and found to be in working order. Thanks @encetamasb for all the help!


encetamasb:
@Injustin

> Adds the 'webhook' notification plugin. The new webhook plugin has been tested for all types of notifications and found to be in working order. Thanks @encetamasb for all the help!

Thanks for the mention!

There are lot of commits in this PR what's not intended to be there.

If you don't have better idea, then I suggest you to create another PR from scratch.. I know it's annoying, but that's how you can fix this without knowing git more:

1. Backup your changes (copy those ~3 source files somewhere).
2. Checkout your master branch and create a new branch from that, and check that out.
```
git checkout master
git branch feature/webhook-plugin
git checkout feature/webhook-plugin
```
3. Copy back your files.
4. Commit, push and PR that. Don't forget to set "master" as target when creating the PR.
5. Ensure that your patch (diff) contains only intended changes.

PR #758, opened 2023-02-06
Adds Webhook Notification Plugin

lnjustin:
Adds Webhook Notification Plugin. All event types tested and confirmed working.

### DIFF
--- a/backend/notifications/plugins/webhook/__init__.py
+++ b/backend/notifications/plugins/webhook/__init__.py
@@ -1,0 +1,158 @@
+from typing import Dict, Optional, Any
+import logging
+import requests  # type: ignore
+import io
+import os
+import json
+from enum import IntEnum
+import requests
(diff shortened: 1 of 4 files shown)


kennethjiang:
@lnjustin Thank you for this PR!

When I test it on my local dev server, it returns a blank page:

Screen Shot 2023-02-06 at 9 31 48 PM

I suspect this is because the Vue component needs to be named `WebhookPlugin`, but I'm not 100% sure as I'm not familiar with the details of how the plugin frontend works.

If necessary, @dipuzyrev may be able to help you with it.

lnjustin:
> I suspect this is because the Vue component needs to be named `WebhookPlugin`

Yes, that re-naming was it (along with an internal change in the vue file as well). Updated.

kennethjiang:
@lnjustin Sorry for the delay. Thank you so much for this awesome PR!

PR #784, opened 2023-04-21
Fix grammatical error and phrasing

northrup:
Clarifying error message through slight wording change / cleanup.


kennethjiang:
Thanks @northrup

PR #792, opened 2023-05-14
Added docker files for ONNX and Darknet runtimes for x86_64 architecture

e-fominov:
This adds new docker and compose files to make it support ONNX runtime on x86_64 with CUDA and CPU.
Also compiles the latest darknet when building the containers.

##How to start:
### CPU + Darknet (Default):
```docker-compose up```

### GPU + Darknet:
```docker-compose -f docker-compose.yml -f docker-compose.x86.darknet.gpu.yml up```

### CPU + ONNX:
```docker-compose -f docker-compose.yml -f docker-compose.x86.onnx.cpu.yml up```

### GPU + ONNX:
```docker-compose -f docker-compose.yml -f docker-compose.x86.onnx.gpu.yml up```

If you want to make testing of containers, add an additional `-f docker-compose.debug.yml` that will open 3333 port on a host mapping the ml_api container. Ex: ```docker-compose -f docker-compose.yml -f docker-compose.x86.darknet.gpu.yml -f docker-compose.debug.yml up```

Then run 
```curl http://localhost:3333/p/?img=https://github.com/TheSpaghettiDetective/obico-server/blob/master/ml_api/test_data/1.jpg?raw=true```
 and you should see some detection coordinates in the output

Docker should also have nvidia-gpu support enabled. [Here is how to install](https://github.com/NVIDIA/nvidia-container-runtime#installation)

This PR uses not the latest images, but ones that should be compatible with K40

### CODE REVIEW

File: docker-compose.x86.darknet.cpu.yml [hunk 1]
```diff
@@ -0,0 +1,13 @@
+version: '2.4'
+
+services:
+  ml_api:
+    image: thespaghettydetective/ml_api_x86_64_darknet_cpu
```
kennethjiang, line 5: Can you also add the relevant files for building these images? So far we have made sure to check in all code and script necessary to recreate the entire cluster from public docker images.

kennethjiang: Another thing to remember is we want, if possible, all docker images to be multi-architecture. Please refer to https://github.com/TheSpaghettiDetective/obico-server/blob/release/scripts/build_dockerfile_web_base.sh for how this can be done.

docker-compose.x86.darknet.gpu.yml
```diff
@@ -0,0 +1,21 @@
+version: '2.4'
+
+services:
+  ml_api:
+    image: thespaghettydetective/ml_api_x86_64_darknet_gpu
```
kennethjiang, line 5: Do you think it's possible that we use the same image for both gpu and cpu? So basically if the GPU hardware is present, the user only need to set "HAS_GPU" to "True", without having to change the image and how the start the server. This will make documentation and supporting a lot easier.

docker-compose.x86.darknet.gpu.yml
```diff
@@ -0,0 +1,21 @@
...
+    build:
+      dockerfile: Dockerfile.x86_64
+      target: ml_api_darknet_gpu
+      context: ml_api
+    # enables GPU access for container
+    deploy:
+      resources:
+        reservations:
```
kennethjiang, line 13: I don't think this is necessary. Our current version doesn't force a resource reservation. If users wants to do it, they can do that via docker-compose.override.yaml. I don't think we want to force this upon users by defau.t

ml_api/lib/backend_darknet.py
```diff
@@ -19,11 +19,17 @@ class BOX(Structure):
...
     _fields_ = [("bbox", BOX),
                 ("classes", c_int),
+                ("best_class_idx", c_int),
                 ("prob", POINTER(c_float)),
                 ("mask", POINTER(c_float)),
                 ("objectness", c_float),
-                ("sort_class", c_int)]
-
+                ("sort_class", c_int),
+                ("uc", POINTER(c_float)),
```
kennethjiang, line 27: Can you provide additional info for this change? I don't know what kind of regression testing I need to run to test this part.

e-fominov: This code is now using latest Darknet version and its interface has changed. You can see it here: https://github.com/AlexeyAB/darknet/blob/master/darknet.py

docker-compose.x86.darknet.gpu.yml
e-fominov: I can remove that. Just need to be sure, that we have a documentation how to run this docker-compose image on CUDA as it may not run without a reservation.

docker-compose.x86.darknet.gpu.yml
kennethjiang: Image size is a good point. The time to launch a new Obico server is already pretty long now. I'll think a little more about options here.

kennethjiang:
Thank you @e-fominov for your prompt action! Will test it later today.

kennethjiang:
@e-fominov I'm closing this PR as I'll continue the work on this PR https://github.com/TheSpaghettiDetective/obico-server/pull/795 and merge it very soon. Thanks again for having done the heavy-lifting to make it happen.

PR #809, opened 2023-07-20
Update ops.md

jamincollins:
For _most_ Linux distributions, if a user is part of the `docker` group, they do not need to use `sudo`. 

Additionally, adding a note about how to clean up older images left behind by upgrades.


kennethjiang:
Thank you @jamincollins!

PR #812, opened 2023-08-20
compose: Improve variable handling on podman

nroach44:
Per the first commit, podman follows the compose spec, 
but the docker-compose.yml did some docker specific things.

This fixes a handful of annoying issues that prevented podman
from being as simple to use with Obico as docker.

**Note**: the `SENTRY_DSN` variable still needs to be set in the .env,
but I'm not sure if I should do that here, or if the code in the
container needs to be fixed.

Prior to these changes, the variables would never be picked up
from .env, and once you added it to the config, the variables
would not work because of the defaulting being non-standard:

```
auth: plain(?,<OBICO>): Username character disallowed by auth_username_chars: 0x24 (username: ${EMAIL_HOST_USER-})
```

(That was only visible after fixing the `EMAIL_HOST` variable)


kennethjiang:
Thank you for submitting this PR. I def want to make it easy for podman users.

However, changing "dotenv.example" -> ".env" will cause the users who have changed ".env" to not be able to `git pull` to update the server. This will be a show-stopper. Any idea for keeping ".env" in ".gitignore"?

nroach44:
It *should* still be in .gitignore, but it'd still be overwritten locally if the remote updates.

One thing we could do is update the setup guide - add a step before run saying something like "copy dotenv.example and adjust to .env and adjust as necessary"

kennethjiang:
> It _should_ still be in .gitignore, but it'd still be overwritten locally if the remote updates.
> 
> One thing we could do is update the setup guide - add a step before run saying something like "copy dotenv.example to .env and adjust as necessary"

Having a file committed git and in .gitignore is a bit counterintuitive and error prone. For instance when/if we updated it to add/remove settings, all existing users who have had it locally modified won't get the change. 

I'd go with changing the setup guide. Need to think a little about where to put it, since I don't want to make it look like a mandatory step.

nroach44:
> Having a file committed git and in .gitignore is a bit counterintuitive and error prone. For instance when/if we updated it to add/remove settings, all existing users who have had it locally modified won't get the change.

Yup, I wasn't suggesting we leave it as is, just commenting that local changes would be ignored by git. Same sort of deal as changes to config files when apt-get tries to update them - it'd error and ask you to fix.

> I'd go with changing the setup guide. Need to think a little about where to put it, since I don't want to make it look like a mandatory step.

I think (not super familiar with the container tech yet) it has to be mandatory, for podman at least. Podman absolutely won't pick up the .env unless it's told to (as that's they way they've interpreted the spec, see https://github.com/containers/podman-compose/issues/458#issuecomment-1076257345 )

That could just be a "if you are running podman, copy dotenv..."

kennethjiang:
Sorry for the delay. Have been busy with Celestrius.

> That could just be a "if you are running podman, copy dotenv..."

The default .env is empty so I'm not sure why it's necessary for pdoman. Which env var are necessary for podman to run?

nroach44:
I believe at the least `SENTRY_DSN` causes issues if it's not defined.

kennethjiang:
> I believe at the least `SENTRY_DSN` causes issues if it's not defined.

I'm pretty sure missing `SENTRY_DSN` wouldn't be an issue as I never have it defined for my local dev (I don't want my dev env errors to contaminate the sentry issue repo).

Did you run into any errors that seem to be caused by missing `SENTRY_DSN`?

kennethjiang:
@nroach44 I'm closing this PR assuming you are no longer available to finish it. Feel free to open a new one if you want to continue.

PR #817, opened 2023-09-10
Fix filament_used for Octoprint API

puterboy:
Added code to retrieve filament_used for Octoprint API.
Not sure if the original line:
```        filament_used = printer_status.get('progress', {}).get('filamentUsed')```
is ever correct for Octoprint as filament usage is part of the 'job' stanza.
So I kept that line in.
Regardless, if that line fails to retrieve the filament_used, then the lines I added properly query the Octoprint API to get the filament usage.

### DIFF
--- a/backend/api/octoprint_messages.py
+++ b/backend/api/octoprint_messages.py
@@ -138,6 +138,11 @@ def update_print_stats_if_needed(printer_status, _print):
 
         completion = printer_status.get('progress', {}).get('completion')
         filament_used = printer_status.get('progress', {}).get('filamentUsed')
+        if filament_used is None: #Try Octoprint api
+            for key in printer_status.get('job', {}).get('filament'):
+                tool_used =  printer_status.get('job', {}).get('filament').get(key).get('length')
+                if tool_used is not None:
+                    filament_used = (filament_used or 0) + tool_used
 
         if filament_used is None and completion is not None and _print.g_code_file and _print.g_code_file.filament_total:
             if completion == 0 and print_time and _print.g_code_file.estimated_time: # Old moonraker-obico version sends completion: 0.0 when print ends. We estimate it using print time

### CODE REVIEW

**Comment by kennethjiang** [comment]
I think a better solution is to change OctoPrint-Obico to be consistent with moonraker-obico: https://github.com/TheSpaghettiDetective/moonraker-obico/blob/42ae428103982908424bb1245609b44b48e52566/moonraker_obico/printer.py#L205

**Comment by puterboy** [comment]
Although my code knowledge is very basic and I have never used Moonraker, from looking at the code, it seems that the OctoPrint-Obico code is basically just doing exactly what the Octoprint API does automatically without Obico -- except that filament_used is under 'Progress" instead of "Job" and that the Octoprint API gives the filament by tool.

I wouldn't be surprised if the Moonraker code was put there to shadow the Octoprint API.

In any case the lines that I added:
```       if filament_used is None: #Try Octoprint api
            for key in printer_status.get('job', {}).get('filament'):
                tool_used =  printer_status.get('job', {}).get('filament').get(key).get('length')
                if tool_used is not None:
                    filament_used = (filament_used or 0) + tool_used
```
do exactly what the preceding line does for Moonraker, using the same routines
```        filament_used = printer_status.get('progress', {}).get('filamentUsed')
```
except that my poor coding skills took 4 lines to sum over the tool elements that Octoprint breaks out.

I'm not sure why one would want to duplicate the already existing Octoprint API structure in Octoprint-Obico.

As I pointed out in my very first post, the Octoprint API is as follows:
```{
  "job": {
    "file": {
      "name": "whistle_v2.gcode",
      "origin": "local",
      "size": 1468987,
      "date": 1378847754
    },
    "estimatedPrintTime": 8811,
    "filament": {
      "tool0": {
        "length": 810,
        "volume": 5.36
      }
    }
  },
  "progress": {
    "completion": 0.2298468264184775,
    "filepos": 337942,
    "printTime": 276,
    "printTimeLeft": 912
  },
  "state": "Printing"
}
 
Of course, I'm sure my pidgin python code can be tightened up...

**Comment by puterboy** [comment]
I should also note that filament_used is broken in other places too for Octoprint.
First of all, the "estimation code" in the lines following the above fail to work for Octoprint:
```if filament_used is None and completion is not None and _print.g_code_file and _print.g_code_file.filament_total:
            if completion == 0 and print_time and _print.g_code_file.estimated_time: # Old moonraker-obico version sends completion: 0.0 when print ends. We estimate it using print time
                completion = print_time / _print.g_code_file.estimated_time
            filament_used = _print.g_code_file.filament_total * completion / 100.0
```
Also, the following code from normalizers.js doesn't seem to work for Octoprint:
``` // Normalize metadata
  gcode.metadata = {}

  if (gcode.metadata_json) {
    // obico file with metadata
    gcode.metadata = JSON.parse(gcode.metadata_json)
  } else if (gcode.analysis) {
    // octoprint file
    gcode.metadata.object_height = gcode.analysis.dimensions?.height
    gcode.metadata.estimated_time = gcode.analysis.estimatedPrintTime

    let filament_total
    if (gcode.analysis.filament) {
      const tools = Object.keys(gcode.analysis.filament)
      if (tools.length) {
        filament_total = 0
        tools.forEach((key) => {
          filament_total += gcode.analysis.filament[key].length
        })
      }
    }
    gcode.metadata.filament_total = filament_total
  } else {
    // either obico file without metadata or klipper file
    gcodeMetadata.forEach((v) => {
      if (gcode[v.name]) {
        gcode.metadata[v.name] = gcode[v.name]
      }
    })
  }
```
I really could use some help understanding the code so I can fix it...

kennethjiang:
At the beginning we tried to let moonraker-obico mirror octoprint's api. But now that we have started to strike a balance between them so that none of them needs to bend backward to accommondate the other. filament_used is one example.

The design here is to have the agents send `progress/filamentUsed` if it's available. Otherwise, the server will estimate it. We haven't had a chance to implement this in OctoPrint-Obico. So the right way to implement your PR is to implement this part in OctoPrint-Obico.

puterboy:
OK - but that would seem like a lot of coding and I am not a real programmer (and certainly not in python) so it's beyond my skillset to create structured code like that (I'm ok at fixing some bags or hacking together a script).
Meanwhile, Octoprint clearly won't work here. So up to you whether to settle in the short term for a a couple of line to fix or to leave it broken until some takes it upon him/herself to rewrite the octoprint_obico code.

puterboy:
If this proposed fix is not going to be implemented, should I file this as a bug report because "filament_used" is totally broken for all Octoprint users based on my understanding of the code.

kennethjiang:
> If this proposed fix is not going to be implemented, should I file this as a bug report because "filament_used" is totally broken for all Octoprint users based on my understanding of the code.

Yes please open a bug report in the repo OctoPrint-Obico so that we can prioritize it accordingly.

kennethjiang:
Closing as we don't have a good solution here.

PR #821, opened 2023-09-30
[WIP] Addition of Helm Charts For K8S

Fenikkusu:
I no longer actively use Docker Compose, and Haven't for a While. I instead heavily use Kubernetes and Helm for almost all my projects and have a K8S cluster setup in-house for many different projects. I decided I could host Obico on my in-house cluster, but didn't see it readily available via helm.

This PR attempts to bridge the gap to K8S by introducing a `helm` directory into the project that contains all the necessary files for running Obico via Helm/K8S.

As this is the first time I'm working with Obico directly, I'm not entirely positive everything is working correctly yet, and there is still a good bit of documentation to complete. At the very least, `it works on my machine` in that I can get the home screen and login. I will continue more in-depth setup/testing later, as I'm brain-dead. I wanted to at least go ahead and get this out there for anyone else to start looking at.

 In theory, so long as you have a linux machine, you should be able to test this simply by running `make` from the `helm` directory (Run it twice the first time).

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -265,7 +265,7 @@ jobs:
         run: |
           set -e
           helm dependency update
-          helm package helm --version=${{ env.GitVersion_FullSemVer }}
+          helm package ./helm/ --version=${{ env.GitVersion_FullSemVer }}
 
       - uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -166,7 +166,7 @@ jobs:
             thespaghettidetective/server:base
             ghcr.io/${{ env.REPOSITORY_LC }}/server:base-1.13
             ghcr.io/${{ env.REPOSITORY_LC }}/server:base
-          context: backend
+          context: .
           file: backend/Dockerfile
           tags: ghcr.io/${{ env.REPOSITORY_LC }}/server:${{ env.GitVersion_FullSemVer_Docker }}
           outputs: type=docker,dest=/tmp/web.image.tar
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -183,3 +183,46 @@ jobs:
           file: ml_api/Dockerfile
           push: true
           tags: ghcr.io/${{ env.REPOSITORY_LC }}/ml:${{ env.GitVersion_FullSemVer_Docker }}
+
+  build-helm-package:
+    runs-on: ubuntu-latest
+    needs:
+      - build-backend-image
+      - build-api-image
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v3
+        with:
+          version: v3.13.0
+
+      - uses: actions/download-artifacts@v3
+        with:
+          name: gitversion.properties
+
+      - name: Load Variables
+        run: cat gitversion.properties >> $GITHUB_ENV
+
+      - name: Update Web Image
+        run: sed -i "s#thespaghettidetective/web#thespaghettidetective/web:${{ env.GitVersion_FullSemVer }}#" helm/values.yaml
+
+      - name: Update ML API Image
+        run: |
+          set -e
+          sed -i "s#thespaghettidetective/ml_api#thespaghettidetective/ml_api:${{ env.GitVersion_FullSemVer }}#" helm/values.yaml
+
+      - name: Update Chart Version
+        run: |
+          set -e
+          sed -i "s#version: 0.0.0#version: ${{ env.GitVersion_FullSemVer }}#" helm/Chart.yaml
+
+      - name: Package Helm
+        run: helm package helm --version=${{ env.GitVersion_FullSemVer }}
+
+      - uses: actions/upload-artifacts@v3
+        with:
+          name: obico-helm-package
+          path: obico-${{ env.GitVersion_FullSemVer }}.tgz
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -262,7 +262,10 @@ jobs:
           sed -i "s#version: 0.0.0#version: ${{ env.GitVersion_FullSemVer }}#" helm/Chart.yaml
 
       - name: Package Helm
-        run: helm package helm --version=${{ env.GitVersion_FullSemVer }}
+        run: |
+          set -e
+          helm dependency install
+          helm package helm --version=${{ env.GitVersion_FullSemVer }}
 
       - uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -79,8 +79,15 @@ jobs:
             ghcr.io/${{ env.REPOSITORY_LC }}/web:base
           context: backend
           file: backend/Dockerfile.base
-          push: true
           tags: ghcr.io/${{ env.REPOSITORY_LC }}/web:base-${{ env.GitVersion_FullSemVer_Docker }}
+          outputs: type=docker,dest=/tmp/web-base.image.tar
+          push: false
+
+      - name: Upload Image
+        uses: actions/upload-artifact@v3
+        with:
+          name: web-base.image.tar
+          path: /tmp/web-base.image.tar
 
   build-base-ml-image:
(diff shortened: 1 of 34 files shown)


kennethjiang:
Helm will be a good addition to different ways to run obico-server.

Can you add a guide (markdown format) to https://github.com/TheSpaghettiDetective/obico-server/tree/release/website/docs/server-guides/platform-specific? You don't have to care much about the format/style as I'll clean it up. But it'll serve to help me test/understand this PR, as well as a foundation for a guide for k8s users.

Fenikkusu:
The one concern I have right now is that K8S is a pretty advanced subject in its own right. Anyone without K8S knowledge already will have a lot of trouble. For example, the Docker Compose version builds the images on the fly. With K8S, you can't [easily] do this. You have to use an image from a proper registry. Now, if you plan on formally hosting Obico in the docker hub, this isn't that big a deal anymore, since people wouldn't need to build the image themselves. I'm cheating a bit right now in the build by leveraging the docker-compose file to build the image, but it assumes that docker & k8s exist on the same machine with the same internal registry (Like Rancher Desktop does).

kennethjiang:
Hmmm I do agree this sounds a bit hacky.

How about separating this PR into 2: the first one addresses building the server docker image and pushing it to ghcr.io. And the 2nd is to create a helm, which should be relatively straightforward.

What do you think?

Fenikkusu:
I think that's reasonable. Right now, this PR as it stands is ultimately that second PR already. The only issue is that I did add a Makefile adjustment to 'build' the docker images using docker-compose, but the more I think about it, the less that is going to work, so it would be better to formally have the first PR completed and adjusting this one to use the form docker images.

kennethjiang:
It looks like https://github.com/TheSpaghettiDetective/obico-server/pull/824 worked! We now have packages available at https://github.com/orgs/TheSpaghettiDetective/packages?repo_name=obico-server. Thanks again for your great PR!

Before you move forward with this PR,  I'd like to take a chance to sort out the versioning strategy, so that you don't have to rework your helm chart when/if we change it later. I don't have experience with gittools/gitversion but my best guess is we should go with the "HighestTagBaseVersionStrategy".

Basically what I'm thinking about is we tag new release version, such as "1.1.2", on the "release" branch. Then the package url will be like `ghcr.io/thespaghettidetective/obico-server/ml_api_base:base-1.1.2.4050` instead of `ghcr.io/thespaghettidetective/obico-server/ml_api_base:base-0.1.0.4050`.

I actually think helm chart should be tied to the formal release version, such as "1.1.2", instead of the dynamic version like "1.1.2.4050", for the stability of the releases. But I may have misunderstood how GitVersion is intended to work so I'm open to suggestions.

Fenikkusu:
Historically Speaking, there are a few different paths that can be taken with regard to version. 

First, keep in mind that the Helm Charts have two fields: `version` and `appVersion`. `appVersion` represents the version of the software contained in the charts. This is typically independent of `version`, as `version` usually directly represents the actual version of the charts. In my opinion, it's best to maintain this line of thinking, as it is a better representation of what the chart is since the images formally dictate the version of the application. Using this line of thinking helps clients /users be able to identify when they need to formally update the helm charts versus just formally updating the application. That is to say, K8S will tell them if they need to update the helm charts.

The primary problem with this approach is that it also means that the clients will have to manually monitor when the application itself gets updated. That is to say, if you update the helm charts version whenever the application is updated, then K8S can formally detect the update, giving clients an easy reference to the version. Clients can use the `latest` tag, but it's usually frowned upon in production environments.

The secondary problem with this approach is that in order to keep the version of the helm charts properly maintained/reflected via something like GitVersion, you have to keep the charts in a separate repo. While not inherently an issue, it is something to be aware of.

In the end, it doesn't really matter. I prefer keeping the versions separate, but I've done both formats before.

With regards to GitVersion specifically, it's ultimately just about applying Semantic versioning. Before getting too deep into GitVersion, I would read up on Semantic Versioning if you aren't familiar with it already [https://semver.org]. GitVersion is a fairly straightforward software in that it is used to auto-calculate the version based on the repo tagging and/or the `mode`. For example, it's currently calculating the version as `0.1.0.4050`. Technically, the version is actually `0.1.0+4050`, but the `+` isn't allowed in docker tags, so I changed it to `.` instead. This ultimately means that the 'version' of the software is `0.1.0`, with 4050 pre-release revisions. The second that you formally tag the repo as say `v1.3.0` based on semantic versioning principles, GitVersion will calculate the next version as `1.3.1+1`, if I recall correctly. 

The biggest thing to know about GitVersion is it has 3 modes it uses to calculate the version: `Mainline`, `Continous Delivery`, and `Continous Deployment`. The default is `Continous Delivery`, which the current `0.1.0+4050` comes from. Until you tag the repo, the version doesn't change, only the RC version. In `Continuous Deployment`, everything is fully automated in the sense that it will auto-increment the version for any commits that happen. The final version, `Mainline`, is my personal go-to choice. It is similar to `Continuous Delivery`, but it tries to auto-increment based on merges into the `main` branch. These can all be configured based on the `GitVersion` file in the project.

So, with that all in mind, to directly address your comments, the `4050` is (in this mode) the `pre-release`, so it's not technically something that you have to deal with. It's perfectly reasonable to use just the Major.Minor.Patch format for the version. The real question goes back to if you want to keep it separate from the release or not.

A bit of a long-winded answer, but hopefully that helps.

Fenikkusu:
I should mention, if you tag the repo, GitVersion should automatically tag the images directly as the version you tag it as.

kennethjiang:
Ok. Maybe you should continue this PR based on the images we have in ghcr. Once you are done with this PR, I hopefully will be able to connect the dots

kennethjiang:
Thanks @Fenikkusu . I'll take a look when I get a chance.

Now it seems that every PR will trigger the build steps and cause the checks to fail because of permission issues. Can you help me figure out how to avoid this? I don't think we actually need to build images on every push. Probably just when a tag is pushed/

Fenikkusu:
I'm actually still working on this. I'm trying to avoid the push errors on forks, but not avoid building on every request. The building on every request can be used to establish full CI/CD automated testing, as you will see.

If after all this you want to limit the build to just the release branch or similar, then it's easy to this via an `if` statement on the job.

Fenikkusu:
One thing I will add though, on that same subject: The base images appear to have no major relation to the primary image, so it can be adjusted such that those definitely do not build on every request. Doing this will avoid excess compute times, by only building those when the relative files change. I would however move them to a separate folder called `docker`, as this will make it easier to distinguish the changes required to trigger those builds.

kennethjiang:
> One thing I will add though, on that same subject: The base images appear to have no major relation to the primary image, so it can be adjusted such that those definitely do not build on every request. Doing this will avoid excess compute times, by only building those when the relative files change. I would however move them to a separate folder called `docker`, as this will make it easier to distinguish the changes required to trigger those builds.

Ok let me know when this PR is ready for review.

Yes the original design of the base image is to capture the configurations that are relatively stable, such as system and python packages, so that users have smaller download size everytime they upgrade.

So I agree that we don't need to build base image everytime. Base image can be build manually, as it should happen very infrequently.

kennethjiang:
@Fenikkusu lmk if you are still interested in pursuing this PR. The docker images in githubs are causing some confusion for our users, as some of them are asking how to use them, a question I don't have a clear answer for (and even if I do, I'd like to have it in the documents so that I don't have to answer it over and over).

Also currently all other PRs are being shown as failing checks. This is another part of the confusion.

Fenikkusu:
I've been really busy the last month and haven't had the chance to do anything else with this PR since. I am still wanting to pursue it, I just haven't had the chance.

kennethjiang:
@Fenikkusu I am closing this PR, as well as reverting #824, as they have caused too much confusion.

Feel free to open another PR when/if you have time to tackle this as a whole.

Fenikkusu:
I understand perfectly @kennethjiang . I will get back to it as soon as I can, but I've been having to work 10 or 11+ hours for the past two months and may need to continue to do so for another month or so. Hopefully shorter, but may be a bit before I can come back to it.

While I'm thinking about it, can you tell me what the issues were with #824? I haven't been able to follow anything on this, unfortunately.

kennethjiang:
No problem. Take your time. 10+ hours a day is brutal.

There were 2 problems with #824:

1. Github will show builds failing for any other PR. This confuses the developers who submitted these PRs.
2. Some users noticed the images built with github actions and thought they should use these images, but we don't have documents on how so this confused them too.

PR #823, opened 2023-10-03
Fix email image attachments

### LINKED ISSUE #822: [BUG] email images not working (self-hosted)
## Describe the bug
Email notification images are not working for me (due to Gmail link rewriting). 

This was originally addressed in #43 by attaching the image rather than embedding (if the link was an internal IP address), but that seems to be broken now.

## Potential Fix
I found the following relevant section of code: https://github.com/TheSpaghettiDetective/obico-server/blob/022d84a77527718b11947057b8604ccef03331c7/backend/notifications/plugins/email/__init__.py#L161

There are two problems that prevent this from working as expected:
  - the `requests` module is not imported in this file
  - invalid reference to `context.img_url` (should just be `img_url`)

I am testing a fix now and will submit a PR if it works.

ApexArray:
Fixes #822

### DIFF
--- a/backend/notifications/plugins/email/__init__.py
+++ b/backend/notifications/plugins/email/__init__.py
@@ -3,6 +3,7 @@ import smtplib
 import logging
 import backoff  # type: ignore
 import os
+import requests
 
 from django.conf import settings
 from django.template.base import Template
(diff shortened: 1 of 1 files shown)


kennethjiang:
Thank you @ApexArray for spotting this bug and this awesome PR!

PR #824, opened 2023-10-04
Adding Docker Build Actions

Fenikkusu:
Adding Docker Build Actions

Adds GitHub Actions to Build Images and store them in the GitHub Container Registry

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -1,0 +1,185 @@
+name: Building Docker Images
+on: [push,pull_request]
+#  pull_request:
+#    branches:
+#      - 'release'
+#      - 'master'
+#  push:
+#    branches:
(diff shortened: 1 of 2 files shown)


Fenikkusu:
I'm going to need some guidance on this: The pipeline is running, but the build of the ML API is failing with an error indicating the specific commit hash is missing. As I'm not very familiar with the inner workings of what Obico is supposed to be, I'm not sure of the best path forward in terms of establishing a replacement hash.

https://github.com/Fenikkusu/obico-server/actions/runs/6411698671/job/17407684714

Fenikkusu:
Okay, I think this should be good now, at least as a first pass.

kennethjiang:
Thank you for the PR. LGTM! And sorry for the delay!

PR #826, opened 2023-10-07
Secure media view

ApexArray:
Enforces authc/authz checks for media files


kennethjiang:
Thank you @ApexArray for this impeccable PR

PR #828, opened 2023-10-18
WIP: Secure media URLs with HMAC signatures

ApexArray:
**Overview**
This PR provides authc/authz for media URLs via HMAC signatures and reverts the previous method enforcing permissions via the URL/file path.

**Signing URLs**
Call the `new_signed_url(url_str)` function to sign a URL. This does the following:
- Generates a `valid_until` timestamp, defaulting to 1 hour in the future (default can be changed via `settings.HMAC_SIGNED_URL_EXPIRATION_HOURS`). Can also be explicitly provided as a function argument.
- Calculates an HMAC digest based on a) URL path and b) valid_until timestamp
- Appends the `valid_until` and `digest` values to the given URL as query-string parameters (ex: `?valid_until=n&digest=xyz`)

**Automatically signing URLs retrieved from the API**
The `SignedHmacUrlSerializer` class inherits from DRF's `SerializerMethodField`. This serializer automatically creates the get_* methods expected by DRF when using the `SerializerMethodField`, in order to sign URLs on-the-fly.

If the URL does not point to this server, or if the path does not start with `settings.MEDIA_URL`, the original URL is returned from the database (preserves cloud behavior).

**Checking if a URL is authorized**
The `HmacSignedUrl` dataclass parses a given URL and provides the `is_authorized()` method. This method returns True if a) the `valid_until` timestamp is still in the future and b) the supplied digest matches the calculated digest of the URL path and timestamp.

**PR Status**
Core functionality is working. Still need to deploy to my live server and test for any misc. calls to media URLs outside the API.


backend/lib/site.py
```diff
@@ -8,3 +8,30 @@ def build_full_url(url):
     domain_name = Site.objects.first().domain
     normalized_url = re.sub(r'^/', '', url)
     return '{}{}/{}'.format(protocol, domain_name, normalized_url)
+
+
+def _get_this_site_url() -> str:
```
kennethjiang, line 13: I think [that method](https://github.com/TheSpaghettiDetective/obico-server/pull/828/files#diff-d05647a89b73f00aa43c2a26ee9c60809918d2dcf575737fc34567318c0db84aR6) can be DRY'd a little more by calling this method.

backend/lib/fs_file_storage.py
```diff
@@ -14,7 +16,11 @@ def save_file_obj(dest_path, file_obj, container, content_type):
         copyfileobj(file_obj, dest_file)
 
     uri = '{}{}/{}'.format(settings.MEDIA_URL, container, dest_path)
-    return settings.INTERNAL_MEDIA_HOST + uri, site.build_full_url(uri)
+    internal_url = new_signed_url(settings.INTERNAL_MEDIA_HOST + uri)
```
kennethjiang, line 19: Was the design to sign the URL when the file is saved, or when the file is to be fetched? I am under the impression that we agreed signing the URL when the file is to be fetched is a more elegant solution. If so, shouldn't this method return the unsigned URL?

backend/config/settings.py
```diff
@@ -33,6 +33,9 @@ def get_bool(key, default):
 SESSION_SAVE_EVERY_REQUEST = True
 SESSION_COOKIE_NAME = 'tsd_sessionid'
 
+# How long before HMAC signed URLs expire
+HMAC_SIGNED_URL_EXPIRATION_HOURS = 1
```
kennethjiang, line 37: We definitely can make expiration a lot longer. Some files, such as the URL embedded in email, may need to be accessed for a few days or even longer.

kennethjiang (commented):
This is a solid PR. I guess the only question I have is if the signing happens when the file is to be saved, or when the file is to be accessed. I am under the impression that we have decided on the latter.

backend/lib/fs_file_storage.py
ApexArray: The [SignedHmacUrlSerializer](https://github.com/TheSpaghettiDetective/obico-server/pull/828/files#diff-03996b8fbe7d2c1ef16ec45987f317b42e0baca33f90b4d988f89cd173684bb8R31-R65) class signs (or re-signs) URLs as they are retrieved from the API.

However, there are several places in the code that save a file object and then pass the returned internal or external URL through to ml_api, tasks, redis, or the web frontend, where it is immediately fetched.
image

Since these URLs don't go through the API/serializer, we need to make sure they are signed before they get passed to other locations. 

I picked this spot because it:
1. ensures the URL will always be signed, even if it doesn't go through the API/serializer/database
2. avoids the need to call `new_signed_url()` in all downstream functions (and any future code changes)
3. if I understand correctly, `fs_file_storage.py` is only loaded for self-hosted instances, so it won't affect the cloud instance.

However, if you think there is an advantage to signing the URLs in the downstream locations then I'm happy to make the changes. Let me know what you think.

backend/lib/site.py
ApexArray: fixed in [f4678db](https://github.com/TheSpaghettiDetective/obico-server/pull/828/commits/f4678db6a21fc07b5370621dd4b3376256a9c465)

kennethjiang: I was thinking about the other way around. But your change is probably more elegant.

backend/lib/fs_file_storage.py
kennethjiang: The original design was:

`internal_url`: the url that can only be accessed by other containers on the same docker network. It shouldn't go through the `build_full_url` method. Instead, it should use `settings.INTERNAL_MEDIA_HOST`, which is `http://web:3334` by default, which is only accessible from other containers, such as "tasks". 

`external_url`: the url that will needs to be accessible from the internet, such as browser or email.

Can you make sure all the downstream code that directly uses `internal_url` all happens within one of the containers, namely "web", "tasks", "ml_api"? If so, then `internal_url` doesn't need to be signed. And I'd like to keep this design consistent with before (and should be simpler too).

backend/config/settings.py
kennethjiang: My suggestion is to have a longer default, such as 2 weeks or 4 weeks. This expiration is essentially a measure against a brute-force method to crack the signature, which is probably highly unlikely given our use case (info is probably not valuable enough to attract professional hackers who can pull this off).

ApexArray:
Reworked PR to only sign URLs on save.

Users will need to run `python manage.py resign_media_urls` in the web container to sign existing URLs.

PR #829, opened 2023-10-26
Django 4.x and other package upgrades *Beta testing needed*

ApexArray:
# Overview

This PR bumps Python and its packages to their most current release, bringing numerous improvements to performance and security. Most notably:
- Python 3.7 -> 3.11
- Django 2.2 -> 4.2
- Django Channels / Daphne 2.5 -> 4.0
- Celery 4 -> 5
- Redis 5.0 -> 7.2

Since I was already refactoring much of the codebase for compatibility with the new versions, I also took the opportunity to make some other improvements:

# Performance optimizations
- Disable SESSION_SAVE_EVERY_REQUEST, and refresh the session every 24 hours instead with custom `RefreshSessionMiddleware` middleware.
- Use [cached sessions](https://docs.djangoproject.com/en/4.2/topics/http/sessions/#using-cached-sessions) with Django's `cached_db` SESSION_ENGINE. 
- Only enable debugging packages if DEBUG=True (nplusone, django debug toolbar)
- Fixed an n+1 query in the PrinterViewSet

# Docker optimizations
- Optimize Dockerfile to take advantage of [layer caching](https://docs.docker.com/build/cache/)
- Use `entrypoint.sh` instead of chained CMD
- Add `docker-compose-build-base.yml` to make it easier to build the base image
- Add `docker-compose-dev.yml` to make it easier to start dev environment
  - Runs dev server using `manage.py runserver` instead of `daphne` (hot reloads server after changing a file)
  - DEBUG=True by default
  - Skip entrypoint.sh for container start (does not run migrate or collecstatic)

# Other changes
- Use pipenv for dependency management
  - Run `cd backend && pipenv lock && pipenv requirements > requirements.txt` to generate a new requirements.txt file
- Optimized GitHub workflows in `build-images.yaml` (these changes made sense to me, but I won't be offended if you don't want to include them in this PR).
- Bump `webpack-bundle-tracker` package in frontend (for compatibility with new `django-webpack-loader` version)
- Optimized Django test suite (tests now complete in ~1s instead of ~35s)
  - Note that some tests are currently failing, but that was the case even before the package upgrades
- Register ALL models in Django admin if DEBUG=True
- Add optional locust configs under `./load_testing/`
- Add `init_db_with_test_data` management command. 
  - This command flushes the current database (deletes all data), and adds a test user and printer (primarily used for setting up a load testing environment)

# Upgrade instructions

**NOTE:** This upgrade should be reversible, but it's still good practice to [backup your database](https://www.obico.io/docs/server-guides/ops/#backup-database) before upgrading.

Simply checkout this branch and run `docker compose up -d`.

# PR Status

I will track any bugs or development tasks here.

I'm using this branch in my live environment and everything seems to be working smoothly, but more testing is needed. Please let me know if you're able to help with beta testing!


ApexArray:
# Load testing results

Adding some quick load test results.

## Test conditions
- All relevant settings and environment details were identical, except for the package versions and refactored code.
- Locust ran on a separate computer to minimize its performance impact
- Initialized fresh database with `init_db_with_test_data` management command
- Tested following endpoints (weighted)
  - 50% `/api/v1/octo/pic/`
  - 16.7% `/api/v1/printers/`
  - 33.3% `/api/v1/printers/1/`
- Strictly measures performance within the web container
  - No active print, so pics were not submitted to ml_api 

## Server info:
- Debian 11 (bullseye) VM
- 2x vCPU E5-2667 v2 @ 3.30GHz
- 24GB RAM
- SSD Storage
- Docker 24.0.1
- PostgreSQL 14.2 (Unix socket)

## Results

### Old
~20 rps
50th percentile latency: 5300ms
95th percentile latency: 5600ms
image


### New
40-50 rps (100%+ increase)
50th percentile latency: 1700ms (~300% faster)
95th percentile latency: 3300ms (~170% faster)
image

## Full test reports
[locust_reports.zip](https://github.com/TheSpaghettiDetective/obico-server/files/13182448/locust_reports.zip)

backend/app/migrations/0074_auto_20231009_2220.py
kennethjiang, line 1: Is this necessary for the django upgrade? I'm definitely the minimalist when it comes to db migrations as they account for probably 90% of the release issues.

backend/app/middleware.py
```diff
@@ -66,66 +66,21 @@ def middleware(request):
...
-                            expires=expires, domain=host,
-                            path=settings.SESSION_COOKIE_PATH,
-                            secure=settings.SESSION_COOKIE_SECURE or None,
-                            httponly=settings.SESSION_COOKIE_HTTPONLY or None,
-                            samesite=settings.SESSION_COOKIE_SAMESITE,
-                        )
-        return response
+        session = request.session
+        if settings.SESSION_COOKIE_REFRESH_INTERVAL >= settings.SESSION_COOKIE_AGE:
+            return ValueError("SESSION_COOKIE_REFRESH_INTERVAL must be less than SESSION_COOKIE_AGE")
```
kennethjiang, line 74: I don't think it's helpful to return an error to the client for a server setting error, as the client can do nothing about it. I'd just remove this check assuming it's the system admin's responsibility to get the settings right (it is).

backend/app/middleware.py
```diff
@@ -66,66 +66,21 @@ def middleware(request):
 
     return middleware
 
-
-# https://stackoverflow.com/questions/2116860/django-session-cookie-domain-with-multiple-domains
-# https://ittone.ma/ittone/django-session_cookie_domain-with-multiple-domains/
-class SessionHostDomainMiddleware(SessionMiddleware):
```
kennethjiang, line 72: I remember this was created to deal with either the webview in the mobile app, or the tunnel. Please test both to make sure they are not broken now it is removed.

kennethjiang (commented):
I realized that this PR would be a lot easier to review if you can limit the scope to the things tightly related to python/django upgrade by:

- Move the changes to how we structure the docker image and dev env to a separate PR.
- Move the front-end related changes to a separate PR, unless they are required by the new django version.

.github/workflows/build-images.yaml
```diff
@@ -141,14 +138,14 @@ jobs:
...
         with:
           cache-from: |
-            thespaghettidetective/server:base-1.13
-            thespaghettidetective/server:base
-            ghcr.io/${{ env.REPOSITORY_LC }}/server:base-1.13
-            ghcr.io/${{ env.REPOSITORY_LC }}/server:base
+            ghcr.io/${{ env.REPOSITORY_LC }}/web-base:${{ env.GitVersion_FullSemVer_Docker }}
           context: backend
           file: backend/Dockerfile
+          build-args: BASE_IMAGE=ghcr.io/${{ env.REPOSITORY_LC }}/web-base:${{ env.GitVersion_FullSemVer_Docker }}
```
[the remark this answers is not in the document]
ApexArray: will move remove GH action changes from this PR

backend/app/migrations/0074_auto_20231009_2220.py
ApexArray: Yes, I believe this migration was necessary for the updated version of `django-safedelete`.

.github/workflows/build-images.yaml
```diff
@@ -72,15 +72,12 @@ jobs:
...
-          cache-from: |
-            thespaghettidetective/web:base-1.13
-            thespaghettidetective/web:base
-            ghcr.io/${{ env.REPOSITORY_LC }}/web:base-1.13
-            ghcr.io/${{ env.REPOSITORY_LC }}/web:base
           context: backend
           file: backend/Dockerfile.base
           push: true
-          tags: ghcr.io/${{ env.REPOSITORY_LC }}/web:base-${{ env.GitVersion_FullSemVer_Docker }}
+          tags: |
```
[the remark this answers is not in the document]
ApexArray: reverted in commit [fd7eca1](https://github.com/TheSpaghettiDetective/obico-server/pull/829/commits/fd7eca16900801dfb347eb60dcc03353c15648ee)

.github/workflows/build-images.yaml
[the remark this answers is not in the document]
ApexArray: reverted in [fd7eca1](https://github.com/TheSpaghettiDetective/obico-server/pull/829/commits/fd7eca16900801dfb347eb60dcc03353c15648ee)

backend/app/middleware.py
ApexArray: Tunnel is working on the web and mobile app 👍

backend/app/middleware.py
ApexArray: fixed in [d368a16](https://github.com/TheSpaghettiDetective/obico-server/pull/829/commits/d368a16fe66a9e97da3865289c9b3aef50407f67)

docker-compose.yml
```diff
@@ -2,9 +2,12 @@ version: '2.4'
 
 x-web-defaults: &web-defaults
   restart: unless-stopped
+  image: ghcr.io/apexarray/obico-server/web:dev-django-4.2
```
kennethjiang, line 5: Having both "image" and "build" here are quite confusing. Is the intention to run the container straight off of the image, or build it?

docker-compose.yml
```diff
@@ -2,9 +2,12 @@ version: '2.4'
...
   build:
     context: backend
     dockerfile: 'Dockerfile'
+    args:
+      BASE_IMAGE: ghcr.io/apexarray/obico-server/web-base:dev-django-4.2
```
kennethjiang, line 10: 2 comments here:

1. How can I build the base image? I'd like to host the base image in the repo of official obico account.
2. "dev-django-4.2"  sounds like it's for development env, not for real deployment?

docker-compose.yml
```diff
@@ -60,13 +63,14 @@ services:
       - "3334:3334"
     depends_on:
       - ml_api
-    command: sh -c 'python manage.py migrate && python manage.py collectstatic -v 2 --noinput && python manage.py runserver --nostatic --noreload 0.0.0.0:3334'
+    command: daphne -b 0.0.0.0 -p 3334 config.routing:application
```
kennethjiang, line 66: Is "daphne" the official django dev server now? last time I check it had a lot of problems.

Also why do we have both "command" here and "ENTRYPOINT" in the Dockerfile? Which one wins? Or both will run, but under different circumstances?

kennethjiang (commented):
A few more comments. More just my questions about how to build my local dev env.

docker-compose.yml
ApexArray: reverted in [fad44bb](https://github.com/TheSpaghettiDetective/obico-server/pull/829/commits/fad44bb4f22e44437cf37544be1198f1f35933ec)

docker-compose.yml
ApexArray: reverted in [fad44bb](https://github.com/TheSpaghettiDetective/obico-server/pull/829/commits/fad44bb4f22e44437cf37544be1198f1f35933ec)

docker-compose.yml
ApexArray: Daphne is the default ASGI server used by Django channels. In fact, the previous version was actually using Daphne under the hood because it [overrides](https://channels.readthedocs.io/en/latest/installation.html) the `manage.py runserver` command. This change simply runs the "production" version of Daphne by default. You can still launch the dev server by running `docker compose -f docker-compose.yml -f docker-compose-dev.yml up -d`

Regarding the entrypoint, this is the preferred method of running "startup" commands during container initialization (migrate and collectstatic in this case). Docker actually passes the "command" through as an argument to the entrypoint. Then the last line of `entrypoint.sh` executes the command without spawning a new PID. The entrypoint can be disabled by setting it to an empty string (for example, in the tasks container)

PR #847, opened 2024-01-04
fix exception in notification plugin stopping further processing

vikwin:
Hi,
I noticed that I did not get failure notifications via pushover anymore and looked into that.

The reason I found was, that another active plugin threw an exception due to an invalid token. Any further plugins were not called anymore, including pushover in my case.

For the general notifications, all plugin exceptions are handled, but that's not the case for the more important failure notifications.
To mitigate the issue, I added the same kind of error handling for failure notification as for general notifications.

Cheers


kennethjiang:
Thank you so much for finding this problem and fixing it for us!

And my apologies for the delay.

PR #848, opened 2024-01-06
Update gmail smtp setup guide

SeriousBug:
Looks like Google updated how the page looks and works, so I updated the text and screenshots.


kennethjiang:
Thank you @SeriousBug

PR #858, opened 2024-01-15
Fix broken/deficient media streaming in backend/app/views/web_views.py

puterboy:
As per ([TheSpaghettiDetective/TSDApp#137](https://github.com/TheSpaghettiDetective/TSDApp/issues/137)), current code in backend/app/views/web_views.py fails to properly implement video streaming -- in particular, it doesn't allow seeking which is required by android mediaplayer used in the Obico mobile app.
I fixed the code to properly implement seeking and streaming adapting a code sample posted on: https://stackoverflow.com/questions/33208849/python-django-streaming-video-mp4-file-using-httpresponse
See the above bug report for more details.
TL;DR summary is:
•	Android mediaplayer (but not exoplayer) do a seek EOF on the media stream and so fail to play timelapses (note: VLC and mplayer have the same problem and can't play the streams when pointed to the media url)
•	The very rudimentary HttpResponse web serving method implemented in the current web_views.py does not allow for seeking (or even real streaming (which probably explains why I have never been able to seek forward/backward in a stream - just pause/resume and rewind to start)
•	Obico cloud works fine since it uses storage.googlesapi.com to properly serve streams
•	Similarly, browsers have no problem downloading and playing the media file as a downloadable file
•	This also explains why the mobile app worked when I hacked it to play a fixed, standard test mp4 file or when I moved the timelapse to my own apache server
•	I suspect that others haven't encountered this problem since they likely use a proxy server (which I do not) which presumablhy 'buffers' the timelapse and allows for seeking (I am just guessing here).
Code has been tested to work with Obic mobile app, Obico web as well as with VLC and mplayer when pointing to the timelapse url's (including digest).
Streaming is more responsive even on web and you can now move forward/backward within the stream using the slider...

### DIFF
--- a/backend/app/views/web_views.py
+++ b/backend/app/views/web_views.py
@@ -2,11 +2,12 @@ import mimetypes
 import os
 from binascii import hexlify
 import re
+from wsgiref.util import FileWrapper
 
 from django.shortcuts import render, redirect
 from django.views import View
 from django.contrib.auth.decorators import login_required
-from django.http import HttpResponse, JsonResponse, HttpResponseForbidden
+from django.http import HttpResponse, JsonResponse, HttpResponseForbidden, StreamingHttpResponse
 from django.contrib import messages
 from django.urls import reverse
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


puterboy:
Note I deleted the previous pull request since it was against the old version of Obico-server... so I updated to be consistent with the changed code...

kennethjiang:
Thank you @puterboy 

Can you re-send this PR against the master branch? 

After that, @ApexArray can you take a look. You are the master mind behind the `serve_jpg_file` function now.

puterboy:
Done!
Kenneth Jiang wrote at about 20:56:57 -0800 on Monday, January 15, 2024:

ApexArray:
Merged in https://github.com/TheSpaghettiDetective/obico-server/commit/dc1e80f59cfbafd884e0aa10196691c26886f73e

Also see https://github.com/TheSpaghettiDetective/obico-server/pull/859

kennethjiang:
Thank you so much @ApexArray

PR #867, opened 2024-01-22
Update klipper-setup with elaboration on configs

HumbleDeer:
Elaborates on the config file locations for Klipper and Moonraker, adds clarification on what Klipper with Moonraker is supposed to mean, and slightly improves the flow of the Klipper install.sh bash script steps.

### DIFF
--- a/website/docs/user-guides/klipper-setup.md
+++ b/website/docs/user-guides/klipper-setup.md
@@ -1,7 +1,7 @@
 ---
 id: klipper-setup
 title: Set up Obico for Klipper
-description: For Klipper/Moonraker/Mainsail/Fluidd users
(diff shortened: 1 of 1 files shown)


HumbleDeer:
Note to self or others: The document outlining the manual code acquirement is both a duplicate of some of the information in *this* guide, as well as missing formatting and links to images.

It may be better to slightly modify the setup script to indicate the user is waiting for pairing, and give the option to hit a key to enter a manual code instead. It would vastly improve the flow of both the setup and the documentation. That is, of course, if this hasn't already been done or implemented as of yet.

kennethjiang:
Thank you @HumbleDeer !

PR #872, opened 2024-01-31
Dockerfile cleanups

nvtkaszpir:


kennethjiang:
Thank you for this PR.

Can you let me know the purpose of this PR? Ideally we have 1 PR for 1 purpose but I'm flexible if you find it easier to combine several related purposes into 1 PR.

By looking at what you have so far, one thing stood out to me is the use of Makefile. In the past other developers have added Makefile to the project but I have found in all circumstances they were not really necessary.

Since this project is  mainly used people (surprisingly) without programming background. So I have tried to keep it simple and tucked away complexity to stay out of the top folder. For instance, the project can still be started by a simple "docker compose up -d" command. Whereas the more complicated tasks, such as pushing to the release branch, is in a shell script in the "scripts" folder.

I don't want you to waste time on things that are not important or compatible with the overall design. So please let me know your purpose and I'll give you my feedback.

nvtkaszpir:
- the change allows to build container images per commit so this could be used in release process and integrated with github workflow later on.
- Makefile allows to build containers in two ways:
  - directly using docker build commands (better when doing single change for specific container
  - docker buildx bake - parallel processing which utilizes docker-build.hcl config and also allows building single target
- docker-compose up still uses old syntax so it works as it was done before

other changes:
- trim down container sizes by merging certain commands together
- less wget spam when downloading data - something like 4 lines instead of pages and pages of dots ;) 
- container labels so it is easier to track container source, which should make it easier to track bugs between released image versions (and thus speed up bisecting bugs)

nvtkaszpir:
We can discuss it on discord. I can split it to smaller PR's, I believe that first priority should be not to rely on docker-compose up and container building on user host - it has some advantages like easier to build from release branch, but makes it impossible to properly release app as packages to some repos such as docker hub, quay.io and others which would allow to scan images for CVE and make it easier to mirror the images to other registries.

Right now docker-build forces some people to build images on their own, and if someone pushes such image then there is rarely any info how it was built and not everyone is eager to check what's inside it (and people really tend to download and run containers without thinking while app in the container runs in privileged mode) so it is a potential attack vector with very specific userbase.

kennethjiang:
i'm fine with using discord to discuss. Not a big fan of going back and forth on github either.

kennethjiang:
> * the change allows to build container images per commit so this could be used in release process and integrated with github workflow later on.
> 
>     * Makefile allows to build containers in two ways:
>       
>       * directly using docker build commands (better when doing single change for specific container
>       * docker buildx bake - parallel processing which utilizes docker-build.hcl config and also allows building single target
> 
>     * docker-compose up still uses old syntax so it works as it was done before
> 
> 
> other changes:
> 
>     * add hadolint to makefiles to lint Dockerfiles https://github.com/hadolint/hadolint
> 
>     * add ARG to Dockerfiles to make it easier to pass certain arguments such as source base image or commit/repo revision
> 
>     * container labels so it is easier to track container source, which should make it easier to track bugs between released image versions (and thus speed up bisecting bugs)
> 
>     * trim down container sizes by merging certain commands together, by merging commands or shallow git clones + remove useless files (could be trimmed even further with adjusting .dockerignore and cleaning up darknet dir even more)
> 
>     * install packages without cache, pin certain package versions
> 
>     * compile darknet a bit faster using number of cpus from the host
> 
>     * less wget spam when downloading data - something like 4 lines instead of pages and pages of dots ;)
> 
>     * use COPY instead of deprecated ADD

Sorry for whatever reason I missed this comment.

I want you to re-consider the use of Makefile. I know that it's a favored tool for some developers. But for our project, it's equivalent of a major retooling of our dev process and hence needs to be justified by significant, tangbile benefits. 

Also about replacing docker compose with docker: is the justification that it'll be faster for the first build? Or is it because docker compose has some intrinsic shortcoming that I'm not aware of?

Let's break it down into multiple PRs, one for each unique purpose/benefit, so that we can have coherent discussion on each of them. And yes discord is preferred if you want to fast-track your PRs. I'm checking github only a couple times a week.

PR #876, opened 2024-02-09
Add CSRF_TRUSTED_ORIGINS as optional environment variable

tefracky:
I had an issue using obico behind a cloudflare tunnel. Hence, CSRF_TRUSTED_ORIGINS must be set in the django config (https://docs.djangoproject.com/en/5.0/ref/settings/#csrf-trusted-origins). This PR adds an optional parameter to the .env to set theses trusted origins.

```
Forbidden (403)
CSRF verification failed. Request aborted.

Help
Reason given for failure:

    Origin checking failed - https://obico.rasgi.de does not match any trusted origins.
    
In general, this can occur when there is a genuine Cross Site Request Forgery, or when [Django’s CSRF mechanism](https://docs.djangoproject.com/en/4.0/ref/csrf/) has not been used correctly. For POST forms, you need to ensure:

Your browser is accepting cookies.
The view function passes a request to the template’s [render](https://docs.djangoproject.com/en/dev/topics/templates/#django.template.backends.base.Template.render) method.
In the template, there is a {% csrf_token %} template tag inside each POST form that targets an internal URL.
If you are not using CsrfViewMiddleware, then you must use csrf_protect on any views that use the csrf_token template tag, as well as those that accept the POST data.
The form has a valid CSRF token. After logging in in another browser tab or hitting the back button after a login, you may need to reload the page with the form, because the token is rotated after a login.
You’re seeing the help section of this page because you have DEBUG = True in your Django settings file. Change that to False, and only the initial error message will be displayed.

You can customize this page using the CSRF_FAILURE_VIEW setting.
```

image

### DIFF
--- a/backend/config/settings.py
+++ b/backend/config/settings.py
@@ -475,6 +475,8 @@ NOTIFICATION_PLUGIN_DIRS = [
 
 ADMIN_IP_WHITELIST = json.loads(os.environ.get('ADMIN_IP_WHITELIST') or '[]')
 
+CSRF_TRUSTED_ORIGINS = json.loads(os.environ.get('CSRF_TRUSTED_ORIGINS') or '[]')
+
 # This line prevents warning messages after 3.2
 # https://docs.djangoproject.com/en/4.0/releases/3.2/#customizing-type-of-auto-created-primary-keys
 DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,7 @@ x-web-defaults: &web-defaults
     ADMIN_IP_WHITELIST: '${ADMIN_IP_WHITELIST-}'
     SITE_USES_HTTPS: '${SITE_USES_HTTPS-False}'
     SITE_IS_PUBLIC: '${SITE_IS_PUBLIC-False}'
+    CSRF_TRUSTED_ORIGINS: '${CSRF_TRUSTED_ORIGINS-}'
     SOCIAL_LOGIN: '${SOCIAL_LOGIN-False}'
     REDIS_URL: '${REDIS_URL-redis://redis:6379}'
     DATABASE_URL: '${DATABASE_URL-sqlite:////app/db.sqlite3}'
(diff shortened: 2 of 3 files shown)


kennethjiang:
Thank you so submitting this PR!

I think I have seen such discussions in our discord server. However, it'll be great if you can list the steps for me to reproduce the CSRF error, so that I can test your PR.

tefracky:
I use a Nginx reverse proxy to access obico. Nginx is accessed via a Cloudflare tunnel and Cloudflare provides the SSL certificate. If I want to access Obico via the public domain --> Cloudflare --> Nginx --> Obico, the CSRF error occurs after login.

kennethjiang:
> I use a Nginx reverse proxy to access obico. Nginx is accessed via a Cloudflare tunnel and Cloudflare provides the SSL certificate. If I want to access Obico via the public domain --> Cloudflare --> Nginx --> Obico, the CSRF error occurs after login.
> 
> I think, this is the same issue: https://discord.com/channels/614543405724205137/648759742881071124/1199353895688880128

What you said here is not steps to reproduce. Your change is simple so maybe I can just blindly merge your change without  reproducing the problem, understanding the problem, and testing the solution. But I have also been burned many times in my 20 years of programming that the "safest" change can deliver surprise if not tested.

Can you think of a reasonable way for me to reproduce the problem on my end?

tefracky:
I don't know, what is the minimum requirement to repoduce the error. However, you can reproduce it by using the Cloudflare Tunnel/Nginx reverse proxy setup as described above. The user in the discord link uses HAproxy and have the same problem. So I think, the reverse proxy is the critical point.
I am also a software developer and I also think, that it would be better if you can reproduce the problem. But unfortunately, I don't know an easy way to to this.

I use this pr branch in my setup currently and all works fine.

kennethjiang:
I tested our cloud set up and it was not impacted. So I went ahead and merged it without reproducing the problem with the full cloudfare setup. 

It'll be great if you can write a short guide to describe the problem and how to solve it with this option in this folder: https://github.com/TheSpaghettiDetective/obico-server/tree/release/website/docs/server-guides/advanced

This will help Google index the keywords so that the users who run into the same issues in the future can jump straight to your solution. But this is optional so if you don't feel like writing docs I completely understand. :)

PR #895, opened 2024-03-15
feat: i18n config test

kui-wang-dada:


kennethjiang (commented):
A question: When we delete a block of code, or a method, or a file in vue, how do we make sure the related translations are also deleted?

For instance, please go through the process of deleting obico-server/frontend/src/components/printer-control/TerminalWidget 2.vue and making sure all obsolete i18n entries are also deleted. Take notes of the steps you go through and share with me

PR #902, opened 2024-03-23
feat: language follow django follow app

kui-wang-dada:


frontend/src/i18n/i18n.js
```diff
@@ -54,12 +45,12 @@ const resources = {
...
     resources,
-    lng: `${language}_${currentSyndicate}`,
-    fallbackLng: `${language}_base`,
+    lng: `${currentLanguage}_${currentSyndicate}`,
+    fallbackLng: `${currentLanguage}_base`,
     interpolation: {
       escapeValue: false,
     },
-    debug: true,
+    debug: false,
```
kennethjiang, line 53: We should leave debug "true" so that it'll prompt us when we miss a key in the translation files.

backend/app/context_processors.py
```diff
@@ -28,5 +28,15 @@ def additional_settings_export(request):
     # per-request syndicate overrides the global setting. This is so that Mintion users can use Obico cloud but see their own theme.
     syndicate_header = request.META.get('HTTP_X_OBICO_SYNDICATE', None)
     if syndicate_header:
-        settings_dict['syndicate'] = {"provider": syndicate_header}
+        settings_dict["syndicate"] = {"provider": syndicate_header,"brand_name": brand_name}
+
+    accept_language_header = request.META.get("HTTP_ACCEPT_LANGUAGE", "")
```
kennethjiang, line 33: Like I commented in TSDApp PR, please check if we can just use Accept-Language to pass the language selection to backennd

frontend/src/i18n/i18n.js
kui-wang-dada: All keys are missing

PR #934, opened 2024-04-24
Added Bearer Token Authentication

andreasvikke:
This Pull requests adds Bearer Token Authentication to the API by utilising the `rest_framework.authtoken` application for Django.

The standard `TokenAuthentication` is using the keyword `Token` for authentication, i ahve therfore extended teh class and changed the keyword to `Bearer`.

My inteand to add this is for other applications to be able to communicate better with the ocibo-server api. Programs like [OrcaSlicer](https://github.com/SoftFever/OrcaSlicer) can only communicate with Bearer on the API.

## Showcase
Django
API

### DIFF
--- a/backend/api/authentication.py
+++ b/backend/api/authentication.py
@@ -19,6 +19,8 @@ class PrinterAuthentication(TokenAuthentication):
 
 def TokenAuthMiddlewareStack(inner): return AuthMiddlewareStack(inner)
 
+class BearerAuthentication(TokenAuthentication):
+    keyword = 'Bearer'
 
 class CsrfExemptSessionAuthentication(SessionAuthentication):
 
--- a/backend/api/viewsets.py
+++ b/backend/api/viewsets.py
@@ -29,7 +29,7 @@ from django.db.models.functions import TruncDay
 from django.db.models import Sum, Max, Count, fields, Case, Value, When
 
 from .utils import report_validationerror
-from .authentication import CsrfExemptSessionAuthentication
+from .authentication import CsrfExemptSessionAuthentication, BearerAuthentication
 from app.models import (
     User, Print, Printer, GCodeFile, PrintShotFeedback, PrinterPrediction, MobileDevice, OneTimeVerificationCode,
     SharedResource, OctoPrintTunnel, calc_normalized_p, NotificationSetting, PrinterEvent, GCodeFolder)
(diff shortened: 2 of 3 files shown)


kennethjiang:
I'm wondering why this is necessary, as obico server is already able to talk to OrcaSlicer. 

@anjingxw can you provide the link to the code about how OrcaSlicer authenticates with obico server?

andreasvikke:
@kennethjiang, after trying to set up OrcaSlicer with my local obico-server running in docker, I noticed the log in screen just showed Blank White Screen, or rather the [o/authorize?...](https://github.com/SoftFever/OrcaSlicer/blob/0369a14ab8a9026d3fc3f4bcdd9bd22424ca0ca5/src/slic3r/Utils/Obico.cpp#L56-L60) wasn't implemented like on the Obico Cloud Hosted.
After some investigation, I figured I could provide an API key in the field further down, and that's where I found out that the Django setup on the obico-server didn't support Bearer tokens yet.

The OrcaSlicer authenticates with Obico Server in [Obico::set_auth](https://github.com/SoftFever/OrcaSlicer/blob/0369a14ab8a9026d3fc3f4bcdd9bd22424ca0ca5/src/slic3r/Utils/Obico.cpp#L48-L54)

I may have missed something in the auth against the Obico-server, if I have please let me know! :)

kennethjiang:
> @kennethjiang, after trying to set up OrcaSlicer with my local obico-server running in docker, I noticed the log in screen just showed Blank White Screen, or rather the [o/authorize?...](https://github.com/SoftFever/OrcaSlicer/blob/0369a14ab8a9026d3fc3f4bcdd9bd22424ca0ca5/src/slic3r/Utils/Obico.cpp#L56-L60) wasn't implemented like on the Obico Cloud Hosted. After some investigation, I figured I could provide an API key in the field further down, and that's where I found out that the Django setup on the obico-server didn't support Bearer tokens yet.
> 
> The OrcaSlicer authenticates with Obico Server in [Obico::set_auth](https://github.com/SoftFever/OrcaSlicer/blob/0369a14ab8a9026d3fc3f4bcdd9bd22424ca0ca5/src/slic3r/Utils/Obico.cpp#L48-L54)
> 
> I may have missed something in the auth against the Obico-server, if I have please let me know! :)

I see. I think you were looking at the wrong "token auth". The token we are using in OrcaSlicer is based on oauth. We added it recently and that's probably why your server doesn't have it.

Please switch your server to pre-release and you should have it. https://discord.com/channels/614543405724205137/661368008781987841/1229143203207319583

andreasvikke:
@kennethjiang, thanks very much for the information! I had totally missed the `pre-release` branch and was not on the Discord till now.

I got it all working on the `pre-release` branch, after setting up an oauth application for OrcaSlicer redirecting to `/orca_slicer/authorized/`

PR #939, opened 2024-05-01
Allow relocating media root

jmerdich:
When self hosting, I make backups of the data I cannot recreate-- a lesson I learned the hard way.  Right now, all user files (eg. time lapses) are stored under the hardcoded path `<gitrepo>/backend/static_build/media` which is intermingled with the source. I want to back up the user files and database, but not the source (it's easy to re-clone it and recreate the containers).

The database can already be redirected to a different file with an environment variable for expert users. This trivial change adds a similar MEDIA_ROOT env var for expert users so that all uploaded files can go to a different place if desired.

### DIFF
--- a/backend/config/settings.py
+++ b/backend/config/settings.py
@@ -453,7 +453,7 @@ GOOGLE_APPLICATION_CREDENTIALS = os.environ.get(
     'GOOGLE_APPLICATION_CREDENTIALS')
 
 MEDIA_URL = '/media/'
-MEDIA_ROOT = os.path.join(STATIC_ROOT, 'media')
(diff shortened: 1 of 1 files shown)


kennethjiang:
I thought I have commented on this PR but apparently I haven't...

I think I understand why you want to make media root customizable. But do you think your problem can be solved by linking `./backend/static_build/media` to a folder of your choice?

The reason why I proposed this alternative is it's not obvious to a lot of people that MEDIA_ROOT should be the path in the container, which is different from the one on the host. I feel that having this configurable will only confuse most self-hosted users, and result in a lot of questions about "why is my server failing now after I have changed MEDIA_ROOT?"

kennethjiang:
I'm closing this PR as part of house-keeping. feel free to re-open it if you still want to continue

PR #951, opened 2024-05-14
Update installation documentation to increase clarity

okjustin:
Add clarification on logging in to accounts locally


okjustin:
Apologies, it seems I missed a change I made somewhere, I will make another PR with all of the changes

PR #962, opened 2024-05-24
Webapp domains mapping

alielshimy:


frontend/src/config/syndicateText.js
```diff
@@ -1,14 +1,20 @@
 export const syndicateTextConstant = {
   base:{
     brandName: 'Obico',
+    docRoot: 'https://www.obico.io/docs',
+    domain: 'https://app.obico.io',
   },
   yumi:{
     brandName: 'Yumi',
+    docRoot: 'https://www.obico.io/docs',
```
kennethjiang, line 9: What will be the behavior of the app when the syndicate is yumi or mintion, since they don't have "domain"?

PR #963, opened 2024-05-24
Print Control Multiple Webcam

manjulsigdel:


frontend/src/views/PrinterControlPage.vue
```diff
@@ -112,25 +112,32 @@
...
         <div class="stream-container">
-          <div v-if="webcams.length > 1" class="webcam-selector">
-            <select id="webcam-select" v-model="selectedWebcamIndex" class="custom-select">
-              <option v-for="(webcam, index) in webcams" :key="index" :value="index">
-                {{ webcam.name || '_default_' }}
-              </option>
-            </select>
+          <div class="header-container">
+            <div class="title">{{ $t('Webcam') }}</div>
+            <div class="webcam-options">
```
kennethjiang, line 117: Where is webcam-options defined? If it's not defined, it means it's not used. Just remove it in this case.

frontend/src/views/PrinterControlPage.vue
```diff
@@ -112,25 +112,32 @@
...
+                <b-dropdown-item v-for="(webcam, index) in webcams" :key="index" href="#" @click="chooseWebcam(index)">{{ webcam.name || 'Primary' }}</b-dropdown-item>
+                <b-dropdown-item @click="chooseWebcam('all')">{{ $t('All') }}</b-dropdown-item>
+              </b-dropdown>
+            </div>
+          </div>
+          <div class="webcam-main" :class="{ 'justify-center' : isAllWebcamSelected, 'webcam-more-than-two' : isAllWebcamSelected && webcams.length > 2 }">
+            <div v-for="(webcam, index) in webcams" :key="index" ref="streamInner" class="stream-inner" :class="isAllWebcamSelected ? (isAtleastOnePrinterPortrait ? 'two-webcam-portrait' : 'two-webcam-landscape') : (videoRotationDeg === 90 || videoRotationDeg === 270 ? 'single-webcam-portrait' : '')" v-show="isAllWebcamSelected ? true : (index === selectedWebcamIndex)">
+              <streaming-box
+                :printer="printer"
+                :webrtc="webcam.webrtc"
```
kennethjiang, line 134: If we are passing webcam to stream-box, then there is no need to pass webcam.webrtc to it again, right?

frontend/src/views/PrinterControlPage.vue
```diff
@@ -155,7 +162,6 @@ import FailureDetectionWidget from '@src/components/printer-control/FailureDetec
 import TemperatureWidget from '@src/components/printer-control/TemperatureWidget'
 import PrinterControlWidget from '@src/components/printer-control/PrinterControlWidget'
 import ReorderModal from '@src/components/ReorderModal'
-import { getLocalPref } from '@src/lib/pref'
 import SharePrinter from '@src/components/printers/SharePrinter.vue'
```
kennethjiang, line 165: getLocalPref is a simple function that deals with edge cases such as users not having local storage enabled. And using it will keept it consistent with the rest of the code base. Any reason why we shouldn't use getLocalPref here

frontend/src/views/PrinterControlPage.vue
```diff
@@ -304,6 +315,42 @@ export default {
   },
 
   methods: {
+    calculateAtLeastOneFeedPortrait() {
```
kennethjiang, line 318: Can you make it so that it uses watcher? Calling this method directly in the code is more bug-prone in case when we add more code to handle webcam orientation and forget to call this.

frontend/src/views/PrinterControlPage.vue
manjulsigdel: streaming-box component is being on other places as well, they depend upon webrtc and stream-box is written to take webrtc from before.

PR #1008, opened 2024-08-09
Infills default selection

sajalregmi:


frontend/src/store/sliceSettingsStore.js
```diff
@@ -5,14 +5,13 @@ const state = {
   ScalebottomSheetOpen: false,
   TranslatebottomSheetOpen: false,
   PrintProfilebottomSheetOpen: false,
-  MachineSelectionOpen: false,
```
kennethjiang, line 8: You reverted a lot of changes I made. Was that on purpose?

sajalregmi: the only change I intended to do was changing the mutation SET_PROFILE_PRESET

sajalregmi: export const fuzzySkinList = ['None', 'Contour', 'Contour and hole', 'All walls'] I saw these list in the code and did it like that.  also the image that Neil said it Liked that. I am not telling I am wrong, I am just telling why I did like that.
image-2

PR #1014, opened 2024-08-14
Slice setting UI 4

manjulsigdel:
- profile overwrites filled with print presets

### DIFF
--- a/frontend/src/store/sliceSettingsStore.js
+++ b/frontend/src/store/sliceSettingsStore.js
@@ -122,6 +122,7 @@ const mutations = {
       internal_solid_infill_pattern: 'monotonic',
       brim_type: 'auto_brim',
      };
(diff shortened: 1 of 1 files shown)


frontend/src/store/sliceSettingsStore.js
```diff
@@ -122,6 +122,7 @@ const mutations = {
       internal_solid_infill_pattern: 'monotonic',
       brim_type: 'auto_brim',
      };
+     state.profileOverwrites = { ...state.profileOverwrites, ...profilePreset }
```
kennethjiang, line 125: Good catch. But I think we should send profileOverwrites to {}.

manjulsigdel: I think this is needed for filling the inputs of printer profile preset. This is for the new way of getting/setting the preset data.

PR #1017, opened 2024-08-16
Changed default brim type to Auto

sajalregmi:


frontend/src/store/sliceSettingsStore.js
```diff
@@ -120,7 +120,7 @@ const mutations = {
     state.profilePreset = { ...profilePreset,
       fuzzy_skin: 'none',
       internal_solid_infill_pattern: 'monotonic',
-      brim_type: 'auto_brim',
```
kennethjiang, line 123: I believe "Auto" is an invalid value. where did you get it from?

PR #1046, opened 2024-11-20
Update docker-compose.yml by adding healthchecks

PiDroid-B:
Hi,

I added healtcheck in docker-compose.yaml
You can take it if you want,

br


kennethjiang:
Thank you for the PR. A few things:

- Can you also update the doc so that the users will know how to use healthchecks to help them run their self-hosted server?
- Can you change the PR or open a new one so that it's based on the `master` branch, not the `release` branch?
- `web` container also has the `/hc/` health check end point. please use that.

PiDroid-B:
i'll do that :+1:

PR #1057, opened 2025-02-03
Update reverse-proxy.md with IIS example

MordantWastrel:
Draft solution for IIS reverse proxy and workaround for Obico to issue cookies for the correct domain

### DIFF
--- a/website/docs/server-guides/advanced/reverse-proxy.md
+++ b/website/docs/server-guides/advanced/reverse-proxy.md
@@ -144,3 +144,69 @@ server {
 1. Restart the Obico Server with `docker compose restart`
 
 1. You should now be able to browse to `spaghetti.your.domain`
+
+## IIS {#IIS}
+
+IIS & Windows can also reverse proxy to Obico. See the below example `web.config` definition using `obico.mydomain.com` as the external address and `192.168.1.100:3334` as the internal address; replace with the appropriate values. 
+
+```
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <system.webServer>
+        <rewrite>
+            <rules>
+                <rule name="ReverseProxyInboundRule1" stopProcessing="true">
+                    <match url="(.*)" />
+                    <action type="Rewrite" url="http://192.168.1.100:3334/{R:1}" />
+                    <serverVariables>
+                        <set name="HTTP_X_ORIGINAL_ACCEPT_ENCODING" value="{HTTP_ACCEPT_ENCODING}" />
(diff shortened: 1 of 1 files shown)


website/docs/server-guides/advanced/reverse-proxy.md
```diff
@@ -144,3 +144,69 @@ server {
...
+                    <add name="HTTP_SEC_WEBSOCKET_EXTENSIONS" />
+                </allowedServerVariables>
+            </rewrite>
+        </system.webServer>
+    </location>
+```
+
+Finally, on your Obico server, you will need to specify the fully qualified domain name for which you want Obico to issue cookies. There is likely some mechanism to make this work automatically (as Obico/Django should be able to detect the domain under which it is being accessed and refer to that value) but it is possible that the above IIS proxy configuration is not relaying the necessary informaiton for it to do that. A workaround is to manually specify the FQDN in `/backend/config/settings.py`. Add these two lines in the `SESSION_COOKIE` section near the beginning of the file, replacing `obico.mydomain.com` with your FQDN:
+
+```
```
kennethjiang, line 209: You mean users need to change the source code? I think a better way is to set them in .env and let them propogate into django. Otherwise most users probably won't know how to do it.

MordantWastrel: Not the source code. Just the difference between editing:

`~.env`

and

~./backend/config/settings.py`

Certainly it would be fine to pull this value from the environment as well (as you're already overwriting other settings values). This is my first time with either Obico or Django though, so I don't know where to find 'overwrite these settings.py values from the environment' but if it's just a matter of appending the variable names onto an existing list, that's easily added to this PR.

kennethjiang: By `~./backend/config/settings.py`, do you mean `[obico-server-path]/backend/config/settings.py` ? If so, it's part of the source code.

MordantWastrel: That's correct. 

I agree that an .env solution would be better, but that will have to come from someone who knows how to implement it.

kennethjiang: Ok... I guess I can merge your PR but I will add a warning that this is highly technical and it may prevent you from getting the future updates (git may refuse to update when you have modified the source code). not ideal but probably the best middle ground we can find

MordantWastrel: Do we not know where the code that says 'pull these values from .env' lives because that's a django thing and not an obico thing?

I feel like it's got to be super simple, I just don't know where to look.

Issue #842, opened 2023-11-24
[Feature] please add support for Rock 5 NPU 


Fuckingnameless:
support for NPU or the mali610 gpu

ss8688:
well，I found a ai detector can work on raspberry pi https://github.com/willuhmjs/forgetti. Not sure about rock 5

PR #1081, opened 2025-06-14
Support for Rockchip NPUs

Laikulo:
This adds support for RKNN-based inference of the YOLO2-based model.
RKNN uses its own model format, which (in this case) is made by converting a preexisting ONNX one.

In addition to a new RKNNNet in ml_api, this also includes.

2 scripts for creating RKNN models:
* `convert_to_rknn` converts a single model from ONNX to RKNN
* `make_rknn_images` converts an ONNX model into a series of RKNN models, targeting different parts.

Two new containers:
* `rknn_toolkit`:  Debian-based with the fill rknn toolkit and a copy of `make_rknn_images`, meant for batch-converting easily.
* `base_arm64_rknn`: An arm64/aarch64 base with the RKNNLite runtime, and rknpu2 library preinstalled.


Laikulo:
Of note: this may also address #842, since the RK3588 has the same NPU series. I don't have any boards with that part, I can't confirm.

kennethjiang:
Oh one thing I should have noticed sooner - the PR needs to use "master" branch as the base. We don't directly merge to the "release" branch.

Can you change the base of this PR to "master"? Thank you!

Laikulo:
TODO:
- [ ] Update docs to indicate that the 'docker-compose' in debian/armbian doesn't support the options used to allow RKNN to read firmware info.

Laikulo:
TODO: Fix broken checked_call in rknn (looks like args/kwargs not behaving as expected)
/draft

PR #1117, opened 2026-01-11
Feat: Webcam window displays static image if webcam stream unavailable

BrandonTwede:
Before:
image

After:
ObicoWithWebcam

The device on which I am running Octoprint does not have Janus installed, nor is this a possibility (running on an old Android phone using Octo4A). Because of this, the obico server does not think I have any webcams installed.  This PR makes it so the webcam window displays the most recent cached image from the server if no webcams are available and a cached image is present . When an active print job is running, I noticed this updates about every 10-15 seconds (Currently the user would have to refresh the page to see new images). I would greatly appreciate this being merged in so I can monitor my prints remotely using Obico, as a periodic screenshot is way better than nothing!

I also added a small fix for the image rotation as 90deg rotation was not working at all for the static image. I am unable to test this with the webcam stream as I do not have a working webcam stream, so let me know if you think lines 478-481 will break anything.

### DIFF
--- a/frontend/src/views/SharedPrinterPage.vue
+++ b/frontend/src/views/SharedPrinterPage.vue
@@ -97,7 +97,6 @@ export default {
         onPrinterUpdateReceived: (data) => {
           this.printer = normalizedPrinter(data, this.printer)
           this.loading = false
-          console.log(this.printer?.pic?.img_url)
           if ((this.printer?.settings?.webcams || []).length > 0) {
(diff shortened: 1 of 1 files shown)


BrandonTwede:
@kennethjiang No problem. I had also toyed with having it update the image automatically on an interval but wasn't happy with the code. I might come back to that later.

What is the release cadence for this project? No rush, just wondering when these changes will be live.

kennethjiang:
> @kennethjiang No problem. I had also toyed with having it update the image automatically on an interval but wasn't happy with the code. I might come back to that later.
> 
> What is the release cadence for this project? No rush, just wondering when these changes will be live.

Just released it. You can update your fork from the upstream now.

Issue #1141, opened 2026-06-14
[Feature] Official Helm chart for Kubernetes deployment


lexfrei:
## Is your feature request related to a problem?

Self-hosting Obico is documented around Docker Compose only. Deploying on Kubernetes today means either hand-converting the Compose stack or relying on the third-party community chart (gabe565/charts), which is not affiliated with or supported by this project. There is no official, project-owned way to run Obico on Kubernetes that tracks server releases.

## Describe the solution you'd like

An official Helm chart, maintained in this repo or a dedicated charts repo, covering the full server stack: `web` (Django/daphne), `tasks` (Celery worker + beat), `ml_api`, and `redis`, with `DATABASE_URL` wired for an external/managed Postgres and a documented path for media storage. Versioned alongside server releases so Kubernetes users get a supported deployment instead of reverse-engineering `docker-compose.yml`.

## Describe alternatives you've considered

- The community chart at gabe565/charts — functional, but unofficial and outside the project's control.
- `kompose` / hand-written manifests from `docker-compose.yml` — brittle and duplicated effort for every self-hoster.

## Additional context

I'm happy to contribute the chart and willing to take it on as a maintainer/co-owner, provided there's some support from the core team on the server's runtime expectations (service dependencies, migration/collectstatic ordering, env/config, scaling assumptions). If you're open to it, just tell me the preferred home (this repo vs. a separate charts repo) and I'll open a PR.

kennethjiang:
Thanks, this makes sense. We are open to an official Helm chart, initially as a beta/community-maintained deployment path.

Preferred home: please start with a PR in this repo under `deploy/helm/obico` or `charts/obico`. We can split it into a dedicated charts repo later if publishing/release automation becomes heavy.

For the first version, please scope it to:
- `web`: Django/daphne
- `tasks`: Celery worker + beat
- `ml_api`
- Redis, either bundled or configurable external
- external Postgres via `DATABASE_URL`
- ingress/service config
- migrate and collectstatic as install/upgrade jobs or init flow
- documented required env vars and secrets
- PVC-backed filesystem media storage mounted at `/app/static_build/media`

[...]

PR #1143, opened 2026-07-07
feat: add Helm chart for Kubernetes deployment

lexfrei:
## What

Adds an official way to run the Obico server on Kubernetes, as discussed in #1141. Three parts:

1. **Helm chart** (`charts/obico/`) — deploys the full stack: `web` (Daphne) and `tasks` (Celery worker + beat) in one pod, `ml-api`, and Redis (bundled or external). External PostgreSQL via `DATABASE_URL`, Ingress or Gateway API for the web UI, PVC-backed media at `/app/static_build/media`, migrate and collectstatic as init containers.

2. **Image build workflow** (`.github/workflows/build-images.yaml`) — obico's docker-compose builds the images from source, so there are no ready-to-run images for a chart to consume. This workflow builds `web`, `ml-api-cpu`, and `ml-api-gpu` from the release branch and publishes them to GHCR under the repository owner's namespace, signed with cosign (keyless).

3. **CPU inference image** (`ml_api/Containerfile.cpu`) — a slim ONNX-Runtime-on-CPU image (~1 GB on disk vs ~7–8 GB for the CUDA image). `detection_model.load_net()` already falls back to the ONNX net when darknet is unavailable, so no application change was needed. Most self-hosted deployments run CPU inference, so this is the chart default; the existing CUDA image is the GPU variant, selected with `mlApi.gpu: true`.

## Scope

Matches the v1 list in #1141: web, tasks, ml_api, Redis (bundled or external), external Postgres via `DATABASE_URL`, Ingress + Gateway API, migrate/collectstatic init flow, documented env vars and secrets, and PVC media (ReadWriteOnce default; ReadWriteMany documented for multiple replicas). Out of scope for v1: HA guarantees, autoscaling, bundled Postgres, object storage.

## Notes for review

- **Images become project-owned on merge.** The workflow publishes under `ghcr.io/<repo-owner>/*` and the chart defaults to `ghcr.io/thespaghettidetective/*`, which line up when merged into this org. The images do not exist until the workflow first runs on `release` after merge.
- **The default image tag is the `release` moving tag.** Pin `image.*.tag` to a `sha-<short>` (immutable per commit) or a CalVer for a reproducible deployment.
- **The SQLite default is evaluation-only** (documented); real deployments point `database.url` at an external PostgreSQL. A connection pooler must run in transaction mode because Obico keeps connections open (`CONN_MAX_AGE=600`).
- Security: non-root containers with dropped capabilities, `automountServiceAccountToken: false` on every pod, an unauthenticated bundled Redis intended for in-cluster use only.

## Testing

`helm lint` is clean, 123 `helm-unittest` cases pass, `values.schema.json` validates `values.yaml`, and the README is generated by `helm-docs`. The CPU image was built and run: gunicorn starts and loads the ONNX model on the CPU.

I'm happy to maintain this going forward.

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -1,0 +1,131 @@
+name: Build and publish images
+
+# Obico does not publish ready-to-run server images — docker-compose builds them
+# from source. This workflow builds them from the current source and publishes
+# tagged, signed images to GHCR so Kubernetes/Helm users can consume them. The
+# ml-api service ships in two variants: a slim CPU-only image (default) and a
+# CUDA image for GPU inference.
+
+on:
+  # The release branch is the deploy branch; rebuild whenever it moves.
+  push:
+    branches: [release]
+  workflow_dispatch: {}
+
+concurrency:
+  group: build-images-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  packages: write
+  id-token: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      # Publish each image independently — one failing build must not block the
+      # others.
+      fail-fast: false
+      matrix:
+        include:
+          - name: web
+            context: backend
+            file: backend/Dockerfile
+          - name: ml-api-cpu
+            context: ml_api
+            file: ml_api/Containerfile.cpu
+          - name: ml-api-gpu
+            context: ml_api
+            file: ml_api/Dockerfile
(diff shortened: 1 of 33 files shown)


kennethjiang:
Thanks for the contribution, Aleksei. We verified the documented local Docker Compose startup still works and are merging this now.

kennethjiang:
The Helm chart and image workflow are now on the release branch. Please update your server checkout and test the release branch version when you get a chance.

PR #1148, opened 2026-07-21
ci(workflows): publish the server images for arm64 and verify what ships

lexfrei:
Splitting the work discussed in #1146 into two steps. This one only builds and publishes images: the install and upgrade commands do not change, and the one compose edit is a gunicorn boot deadline rather than a change to what compose builds. Switching compose from building to pulling comes as a separate PR once these images exist, so that one can be checked on every platform against real artifacts instead of promises.

## What changes

`obico-web` and `obico-ml-api-cpu` are now built for `linux/amd64` and `linux/arm64` on native runners and merged into one manifest list per image, signed as a list. Until now every published image was amd64-only, so anyone on arm64 had to build on the machine itself.

New `obico-ml-api-jetson` image (arm64): on `linux/arm64` the ml_api Dockerfile resolves its base manifest to the L4T/JetPack half, so the same file produces the Jetson build. Publishing it under its own name keeps the hardware target visible instead of hiding it in a manifest list's arm64 slot. It builds best-effort, so a broken upstream archive cannot hold up a release, and it is exercised as a CI build only, never started on hardware.

The CUDA `ml-api-gpu` build stays amd64-only, since CUDA is.

Nothing is tagged before it is checked. Every build pushes by digest and the tags go on afterwards, so an image that finishes building but cannot load a model never owns the tag a chart pins. The web image gets an import check for its compiled dependencies, the CPU image is started and asked for a health response, and the assembled manifest list is inspected before publication to confirm it covers the platforms it claims.

Builds no longer trust what they download. Model weights and the Rockchip runtime were fetched with no verification at all, so an error page served with a 200 would ship as a model and the container would die at import with no hint why. Every download now goes through one checked helper: a non-success status fails, a truncated transfer fails, and the bytes have to match a checksum. Writing that helper turned up a real break. The NVIDIA wheel the Jetson base installs sits behind a link that answers 301 with an empty body, so that base had been unbuildable and nothing was rebuilding it to notice.

Every image gains a `<calver>-g<short-sha>` tag, and the published chart gains the same. The bare CalVer is day-granular and moves when two releases land on the same UTC day, which is not something a chart or a compose file can pin to.

The chart side of this (a startup probe for ml-api and docs for the arm64 and Jetson options) ships as its own PR, so the chart review does not have to happen inside this one. One chart line does ride here: the ml-api command override in values gets the same `--timeout=120` the image CMD gains, so a chart deployed from master keeps the worker boot deadline while the chart PR is in flight. That PR supersedes the line with a startup probe sized from the same deadline.

## What stays the same

Docker Compose still builds from source. The one edit there gives gunicorn a 120s boot deadline: the detection model loads at import, inside worker boot, and the 30s default was killing the worker mid-load and respawning it forever. Install and upgrade commands are untouched, and the Helm chart keeps consuming the same images from the same place.

One documentation change does affect what a reader types, and it corrects an instruction that was already wrong: the RK3588 quick start is now two steps, because the base image it starts from is not published and the old one-liner could only ever fail on the pull.

## Notes for review

New GHCR packages are created private, so `obico-ml-api-jetson` needs a one-time flip to public after the first successful build, the same step the existing packages went through.

A second workflow runs on pull requests and pushes nothing: chart lint and unit tests, the two images most people run built on both architectures and exercised by the same checks the release uses, and the three hand-written ML base images compiled. Those bases were built by nobody until now, which is how one of them came to be broken without anyone noticing.

`onnxruntime-gpu` in the CUDA base is pinned at what an unpinned install already resolves to, and that is the only claim the pin makes. From 1.19 these wheels target CUDA 12 while the base is CUDA 11.4, so a session on CUDAExecutionProvider may well fail on real hardware. Importing the module succeeds either way, so no CI job can see that, and darknet is the backend that image actually uses. Answering it needs a GPU. The pin at least makes the next change to this dependency an explicit diff rather than a silent resolve.

## Verification

`actionlint`, `shellcheck`, `helm lint` and `helm unittest` all run in CI now rather than only on a contributor's machine. Two small scripts cover what no linter does: every documented values key reaches the generated table, and each weights download is gated on what its loader can actually read. Both carry their own self-test.

Since the publishing workflow has no pull request trigger, I ran it end to end on a fork rather than guessing: both platform legs of each multi-arch image, both manifest merges, the CUDA build, the Jetson build and the base builds all green, and the published manifests carry the platforms they should.

I then pulled those arm64 images on an Apple Silicon Mac and ran the server against them: web, tasks, ml_api and redis all healthy, ONNX inference loaded and returned a detection, and a click through the dashboard, print history, stats and the printer wizard produced no errors in the browser or the server log.

Two things could not be exercised on a fork, so I reproduced them against a local registry instead: that a chart published under a second, immutable tag resolves through Helm by that version, and that a short SHA which comes out all digits with a leading zero is not a valid semver prerelease, which is why the tag carries a `g` prefix. Five commits in recent history have such a SHA.


kennethjiang:
Thank you for the PR!

In #1146, my primary concerns about adding complexity to the user experience were:

1. Changes to install and upgrade commands or prerequisites.
2. Requiring users to choose between different setup paths.

This PR addresses the first concern well for the default amd64 path. The normal install command remains the same, and upgrades are slightly simpler because users no longer need `--build`. I like that.

However, the second concern remains, and the transition is not automatic for every existing user. Generic arm64 users, Jetson users, and users who build from source now need to know which Compose override applies to them. In particular, existing generic arm64 users who previously used the default command must choose the CPU-image or source-build path, and Jetson users must replace the old runtime-only override with the Jetson Compose file.

Ideally, existing users should transition without needing to find and read a specific document. That appears achievable for the normal amd64 path, but not for these other paths. If automatic transition is not possible, we need a mechanism to present the relevant guidance at the point of transition—for example, detect the architecture or unsupported Compose setup and show a clear error with the exact command the user should run.

I hope this makes sense. I encourage you to put yourself in the shoes of users following different paths—across platforms and as both new and existing users. Make sure you can walk through the experience for each affected user group without encountering any obvious points of confusion. In my experience, users can still be confused even when the friction is not immediately visible to us. If we can already identify areas of confusion, it is almost certain that users will struggle with them.

lexfrei:
Your point about the other paths sent me back to check what the transition actually looks like today, and it turned out I could not make it safe in one PR. So I cut this one down: it now only builds and publishes images, and touches no compose file at all. The install and upgrade flow is byte-for-byte what it is today.

What forced the split: the images the compose file would pin are amd64-only right now, and the multi-arch ones can only exist after this workflow merges and runs. That is a window where an arm64 user is worse off than before, which is exactly the kind of thing you were asking me to look for. Rather than paper over it in a comment, this PR produces the images first.

Phase two then does the compose switch against artifacts that already exist, so I can walk each group through it for real instead of predicting: plain amd64, arm64, GPU, Jetson, RK3588, build-from-source, both as a fresh install and as an existing user upgrading. That is also where the transition mechanism you asked for belongs, and I have it working already: ml_api detects a GPU it cannot use and logs the exact command to fix it, so an existing GPU user who upgrades finds out from their own container log rather than from a doc they have to know exists.

Nothing here changes what a user runs, so this one is judged on the images alone.

kennethjiang:
I assume this PR is not ready for review yet since it's still in the draft state.  Ping me or flip it to ready for review whenever.

lexfrei:
Yes, still draft on purpose. The work grew, so I'm splitting it up for easier merging: this PR is now only the image builds and publishing, the Helm chart changes are moving to a separate chart-only PR, and the compose switch stays as phase two as discussed above. The last full rehearsal on my fork caught an ordering bug in the new signing job, and I want one clean end-to-end run before taking your time. I'll flip it to ready and ping you then.

lexfrei:
@kennethjiang Ready for you now. The full publish flow ran on my fork end to end: multi arch builds, manifest merge, and both signing legs came back green. CI on this branch is green as well. Since your last look the chart half moved out to #1150, so this one stays images and workflows only, plus a single values line that keeps the worker boot deadline for chart users until the chart PR lands.

kennethjiang:
Looking good. Sorry for taking quite long 'cause I need to run tests for all the paths impacted.

kennethjiang:
Thank you for working through this and for the thorough validation.

PR #1149, opened 2026-07-30
Update website dependencies and standalone builds

cwil2072:
## Summary

- update all Docusaurus packages from 3.7.0 to 3.8.1
- migrate @svgr/webpack from 5.5.0 to 8.1.0 and resolve compatible Webpack transitives
- regenerate the npm and Yarn lockfiles, repairing the malformed Yarn v1 lock
- point documentation links and assets maintained outside this checkout to canonical obico.io URLs so clean-checkout builds do not require external blog files

This is a clean, tested alternative to #1075. It intentionally stays on Docusaurus 3.8.1: version 3.10.2 was evaluated but caused a repeatable heap OOM during the documented docs dev-server workflow.

## Validation

- npm ci
- yarn install --frozen-lockfile --non-interactive
- yarn build for all 9 locales
- npm audit --omit=dev: 74 advisories to 50
- full npm audit: 74 advisories to 50
- npm audit fix --package-lock-only --ignore-scripts --dry-run: no additional safe changes

## Known issue

The yarn start command did not become ready within 2m29 and reached approximately 8.6 GB RSS. Untouched master on Docusaurus 3.7 showed the same behavior, reaching approximately 8.4 GB after 2m24, so this appears pre-existing. Docusaurus 3.10.2 showed a worse immediate heap OOM at both 4 GB and 8 GB on Node 24 and also reproduced on Node 22, so it was not retained.

No test or type-check scripts are defined for website.

### DIFF
--- a/website/docs/user-guides/first_layer_ai/nozzle-camera-printer-installation.md
+++ b/website/docs/user-guides/first_layer_ai/nozzle-camera-printer-installation.md
@@ -37,7 +37,7 @@ This guide will show you how to install an endoscope camera on your Prusa Mini 3
 
 Required Components:
 
-![](/img/blogs/prusa-mini-nozzle-camera-setup/001.jpeg)
+![](https://www.obico.io/img/blogs/prusa-mini-nozzle-camera-setup/001.jpeg)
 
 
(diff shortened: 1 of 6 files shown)


kennethjiang:
`website` is connected to our CMS/blog workflow and isn’t intended for self-hosted users to build. Self-hosted users have no practical reason to build or deploy the marketing site or blog.

Is there a use case I’m missing that requires the static website to build cleanly from the `obico-server` repository alone?

PR #1153, opened 2026-08-13
Add rewrite_media_url_host command for repairing media URLs after a server address change

### LINKED ISSUE #1152: [BUG] Changing the server domain breaks all historical timelapses, and resign_media_urls cannot repair them
**Describe the bug**

Media URLs (timelapses, detective/prediction views, poster images, printer event snapshots) are stored in the database as full URLs, including the scheme and host the server had at the time. Example:

```
http://192.168.1.50:3334/media/tsd-timelapses/private/xxx.mp4?digest=...
```

When a self-hosted server's address changes (e.g. moving from a LAN address to a public domain, following the documented process), every media URL recorded before the change still points at the old host. All historical timelapses, detective views, and downloads break. New prints are fine.

The provided repair command cannot fix this. `resign_media_urls` calls `new_signed_url()` in `backend/lib/url_signing.py`, which only replaces the query string (the digest) and keeps the stored scheme and host exactly as they were. After running it, the URLs have a fresh digest but still point at the old, dead host.

The HMAC digest is calculated from the URL path only — the host is not part of the signature. So rewriting the host of a stored URL never invalidates the signature. There is just no code path that does it.

This is a code defect rather than a configuration issue: the new domain is configured correctly (new prints work end to end), and the URL-preserving behavior is visible directly in `new_signed_url()`.

**To Reproduce**

1. Run a self-hosted server at one address and complete a print, producing a timelapse.
2. Change the server's address following the documented domain-change process.
3. Run `python manage.py resign_media_urls`.
4. Open the print from step 1: the timelapse and detective view fail to load, and the download button opens a URL on the old host.
5. Inspect `Print.video_url` in the database: the host is still the old address.

**Screenshots**

N/A — the failure is visible in the stored URLs above.

**Hosting environment:**

 - OS: Ubuntu Server (Docker Compose install)
 - Version: release branch, commit 49c0bc70
 - Hardware: x86 PC

**Additional context**

Every address the server has ever run at leaves behind its own set of broken URLs. My database had three historical hosts — 443 affected rows across the `Print`, `PrinterEvent`, and `PrintShotFeedback` tables.

`GCodeFile` URLs are stored as relative paths (43 rows on my server), so a fix must leave relative URLs untouched.

PR #1153 is attached that adds an optional `--rewrite-host` flag to `resign_media_urls`. Tested against the deployment above: all 443 rows rewritten, relative URLs skipped, historical timelapses and downloads working again.

ddm-j:
Fixes #1152 and the issue referenced in #1129

## What changed

A new management command:

```
python manage.py rewrite_media_url_host --old-origin http://192.168.1.50:3334 --new-origin https://obico.example.com
```

It rewrites the scheme and host of every stored media URL that matches `--old-origin` so it points at `--new-origin` instead. Only URLs on the old origin are touched; URLs on any other origin (for example externally hosted or provider-signed URLs), and relative URLs, are left as they are. The rest of the URL, including the query string and digest, is preserved.

Both origins are required. `--dry-run` reports how many URLs would be rewritten per model without saving anything. The command prints per-model counts either way.

`resign_media_urls` is unchanged in behavior. The loop over media URL fields (and the progress bar it uses) is now shared between the two commands in `lib/media_urls.py`, next to the other shared helpers. Happy to move it if you'd prefer another location.

The server guide gets a short section under "Django Site" explaining when to run the command.

## Why

Media URLs are stored in the database as full URLs, including scheme and host. When a self-hosted server's address changes, all previously stored timelapse, detective, poster and event snapshot URLs keep pointing at the old address and break. `resign_media_urls` cannot repair this: it is for `DJANGO_SECRET_KEY` rotation and only replaces the query string.

The HMAC digest covers the path only, so rewriting the host never invalidates the signature. See the linked issue.

## Review history

The first version of this PR added a `--rewrite-host` flag to `resign_media_urls`. Per review it is now a standalone command that requires both origins and preserves query strings.

## How I tested it

- Unit tests for `rewrite_url_origin()` in `backend/lib/tests.py`: matching origin rewritten with the rest of the URL preserved and the digest still valid, other origins untouched, relative URLs untouched, case-insensitive host and trailing-slash tolerance, invalid origins rejected.
- Command tests in `backend/app/tests.py` via `call_command()`: matching rows rewritten and still valid, other-origin and relative rows untouched, invalid origin rejected before anything is changed, dry run changes nothing.
- A test that `resign_media_urls` still keeps the host and produces a valid digest after the shared-helper refactor.
- Ran the backend suite with `docker compose -f docker-compose.yml -f docker-compose-dev.yml run --rm web python manage.py test`. The 33 errors on clean `master` (test setup creating users without a Syndicate row) are unchanged; all new tests pass.
- I ran the previous version of this change against my own server (443 affected rows across three old addresses); all historical timelapses and detective views work again.

## Notes

- No migrations, no configuration changes.
- One side effect of the shared helper: rows whose URLs come out unchanged are no longer saved (previously `resign_media_urls` saved every row it visited). Stored URLs are byte-for-byte the same either way; only `updated_at` timestamps on unchanged rows are no longer bumped.

### DIFF
--- a/backend/app/management/commands/resign_media_urls.py
+++ b/backend/app/management/commands/resign_media_urls.py
@@ -31,8 +31,17 @@ def print_progress_bar(iteration, total, prefix='Progress:', suffix='Complete', 
 class Command(BaseCommand):
     help = '(re-)signs all media URLs. Must be run once after updating, and any time the Django SECRET_KEY is rotated'
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--rewrite-host',
+            default=None,
+            help='Also rewrite absolute media URLs to point at this host, e.g. "obico.example.com". '
+                 'Use after the server domain has changed. The scheme is taken from the SITE_USES_HTTPS setting. '
+                 'Relative URLs are left untouched.'
+        )
+
     @staticmethod
-    def _resign_urls_on_model(obj: models.Model, url_fields: List[str]):
+    def _resign_urls_on_model(obj: models.Model, url_fields: List[str], rewrite_host=None):
         changed = False
         total_rows = len(obj.objects.all())
         print(f"Resigning {obj.__name__} URLs ({total_rows} rows)...")
@@ -40,7 +49,7 @@ class Command(BaseCommand):
             for url_field in url_fields:
                 url = getattr(row, url_field)
                 if url:
-                    setattr(row, url_field, new_signed_url(url))
+                    setattr(row, url_field, new_signed_url(url, rewrite_host=rewrite_host))
                     changed = True
             if changed:
                 row.save()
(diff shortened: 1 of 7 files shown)


kennethjiang:
Thanks for the thorough report and tests. The problem and need for a repair command make sense.

Could this be implemented as a separate `rewrite_media_url_host` command instead of extending `resign_media_urls`? Re-signing is specifically for `DJANGO_SECRET_KEY` rotation, while changing a host is an independent operation—the signature covers only the path.

I’d also prefer requiring both the old and new origins, and rewriting only URLs matching the old origin. Rewriting every absolute URL could unintentionally modify externally hosted or provider-signed URLs. Preserving their existing query strings would also be safer.

The model/field traversal can still be shared between the two commands to avoid duplication.

ddm-j:
@kennethjiang - sure thing. I was on the fence about submitting this in the first place. Glad you like the idea.

I'll re-tool it later tonight when I get a chance and admmend the PR.

ddm-j:
@kennethjiang  - reverted `resign_media_urls` and added `rewrite_media_url_host` requiring both old/new hosts. Also added `--dry-run` so that users can preview the changes, and updated docs.

Since these functions share some database traversal code, I extracted the shared helpers to `lib/media_urls.py`. Made a slight change to the DB code, only saving changed rows (this is new). Happy to relocate if this doesn't work for you.

kennethjiang:
Thanks for reworking this so thoroughly. The standalone command, explicit old and new origins, dry-run support, and shared traversal make the repair much safer. Merged—thank you for the contribution!

PR #1157, opened 2026-08-18
Server-side TURN configuration for WebRTC webcam streaming

ddm-j:
Follow-up to TheSpaghettiDetective/moonraker-obico#129 and TheSpaghettiDetective/moonraker-obico#123. Per @kennethjiang's direction there, obico-server becomes the single source of truth for TURN configuration. Companion agent PRs (moonraker-obico, OctoPrint-Obico) will consume the new field.
This feature comes in three parts (none of which depend on each other to work):

1. obico-server (this PR)
2. moonraker-obico (TheSpaghettiDetective/moonraker-obico#130)
3. OctoPrint-Obico (TheSpaghettiDetective/OctoPrint-Obico#248)

This is a decently large diff, but the agent diffs will be pretty small in comparison (most of the work is on the server side).

Everything has been designed such that merges/deployment of this feature can happen in any order. Fully backward compatible, and no impact to existing cloud TURN config (see default turn constants in the backend).

## What changed

New `.env` settings (`TURN_SERVER`, `TURN_PORT`, `TURN_TRANSPORTS`, plus one credential mode) describe a TURN server, which feeds through `docker-compose.yml` like the other settings. When set, the server gives the same configuration to both WebRTC endpoints:

- **web frontend** — as `page_context.syndicate.turn`; `webrtc.js` builds the `RTCIceServer` from it.
- **printer agents** — as `printer.turn` in the `GET /api/v1/octo/printer/` response the agents already fetch at startup, so they configure Janus without any local settings (single source of truth).

Two credential modes, exactly one must be configured (startup fails otherwise):

- `TURN_SECRET` — time-limited credentials per the TURN REST API scheme (coturn `use-auth-secret`, also eturnal/Pion/hosted providers). The server mints `<expiry>:<label>` / `base64(HMAC-SHA1)` credentials: 24h for browsers, 1y for agents (both configurable). Recommended and documented first.
- `TURN_USERNAME` + `TURN_CREDENTIAL` — static long-term credentials (coturn `lt-cred-mech`), for TURN servers that only support that.

See **credential exposure** below for considerations on security.

The contract shape is the same in both places:

```json
"turn": {"server": "…", "port": 3478, "transports": ["udp", "tcp"], "username": "…", "credential": "…", "expires_at": 1234567890}
```

`expires_at` is `null` in static mode. Agents can use it to warn before their credential expires.

**Backward compatibility.** When `TURN_SERVER` is unset nothing changes: `turn` is `null`/absent, `webrtc.js` falls back to the existing `syndicate.turn_server` + auth-token behavior, and agents keep their built-in `turn.obico.io` defaults. Old agents ignore the new field. So the Obico Cloud is unaffected unless it opts in.

**Credential exposure.** `page_context` is rendered into every page including the login page, so TURN credentials are only added when the request is authenticated, or when it is a valid printer share page (token verified against `SharedResource`). This is the best I could come up with to minimize the exposure surface area when a user insists on using "static" mode. Tests cover the login page not carrying credentials, and an unknown share token not carrying them. The context processor no longer mutates the module-level `SYNDICATES` dict (it did before, harmlessly, for `name`; per-request credentials made a copy necessary).

Docs: a new server guide, "Set up a TURN server for webcam streaming", with coturn config for both modes and a strongly suggested hardening section (`denied-peer-ip` for private ranges, quotas). Linked from the webcam FAQ and `dotenv.example`. Added some cautions up top indicating that this type of advanced configuration is generally for people exposing their server to the public internet, pointing them to more common troubleshooting for webcams if they find themselves on that page.

## Why

Self-hosted servers have no way to configure a TURN server today: `turn_server` doesn't exist in the open-source `SYNDICATES`, and the agents hard-code `turn.obico.io` with the printer's auth token, which a self-hosted server's token can't satisfy. `ignore_unreachable_ice_server = true` in the generated Janus config then makes Janus continue silently without relay candidates. Result: streaming works on the LAN and fails from mobile networks behind carrier-grade NAT, with nothing in the logs. Details and evidence in moonraker-obico#129.

## Testing

- `lib.tests.TurnConfigTestCase` — helper: unconfigured, static, secret (HMAC checked against an independently computed value), per-label variation.
- `api.tests.OctoPrinterViewTestCase` — agent endpoint: absent when unconfigured, correct shape and expiry, invalid token rejected.
- `app.tests.PageContextTurnTestCase` — login page has no `turn`; authenticated page and valid share page do; unknown share token doesn't; unconfigured is `None`; `SYNDICATES` not mutated.
- Pre-existing test failures carried through, no new failures.
- Startup validation exercised in the web container for: both modes set, neither set, bad transport, bad port.
- End-to-end on my self-hosted stack: obico-server branch + coturn in `use-auth-secret` mode + the moonraker-obico branch on the printer. The server minted a `<expiry>:printer-1` credential, the agent wrote it into `janus.jcfg`, and the coturn log shows `ALLOCATE processed, success` for that username, `CREATE_PERMISSION processed, success`, and non-zero relay usage. Webcam stream verified working from a phone on mobile data. The test also caught a docs bug: the hardening example denied `100.64.0.0/10` (CGNAT), which 403'd the permissions for cellular peers; the docs now say not to deny that range.

## Not in this PR

- Agent PRs for moonraker-obico and OctoPrint-Obico consuming `printer.turn` (in progress).
- Exposing `turn` on a JSON API for native clients (Mobileraker, a native TSDApp path if one exists). If TSDApp is webview-based then it gets it via `page_context` already (assumption).


kennethjiang:
Thanks for the PR. Could you revise the design around these two points?

1. Please make the per-printer TURN information as consistent as possible across agents and the web/mobile frontend, ideally using the same `printer.turn` response shape.

2. Obico Cloud uses per-printer TURN credentials rather than one static credential. We therefore need a configurable implementation that `tsd-enterprise` can override. Please use the existing `FD_MODULE`/`failure_detection` mechanism as the reference: open-source provides the default settings-based implementation, while enterprise selects its own implementation through `settings_ent.py`.

ddm-j:
Working on this when I can, should be able to get an update this weekend.
I'll dig into `tsd-enterprise` and the failure detection module as suggested.

PR #1159, opened 2026-08-24
ci: build ml-api-cpu for linux/arm64 too

hibikipr:
## Summary
`ml-api-cpu` has no GPU/CUDA dependency, and both pinned inference deps (`onnxruntime==1.16.3`, `opencv-python-headless==4.9.0.80`) ship `aarch64` wheels - so there's no reason to limit it to `linux/amd64` the way `ml-api-gpu` (genuinely CUDA/amd64-only) has to be. `web` stays `linux/amd64`-only too - not changed here.

I run `ml-api-cpu` on an arm64 host (Apple Silicon) and had been building it from source locally there instead, since this workflow didn't publish anything that ran natively on it.

## Test plan
- [x] `docker/build-push-action` succeeded for `ml-api-cpu` on `linux/amd64,linux/arm64` from my fork's own run of this exact workflow (QEMU-emulated arm64 leg, same mechanism this PR relies on)
- [x] Verified the published `linux/arm64` manifest is real (`docker manifest inspect` shows both platforms) and functional - pulled and ran it natively on an arm64 host, `/hc/` responds `ok`, gunicorn starts clean

Happy to adjust anything to match how you'd want this maintained.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kennethjiang:
Thanks for this — and thanks for actually verifying the wheels rather than assuming. You're right that `ml-api-cpu` has no reason to be amd64-only, and the dependency audit checks out: `onnxruntime==1.16.3` ships a cp38 `manylinux_2_17_aarch64` wheel, and `opencv-python-headless==4.9.0.80` ships a `cp37-abi3` aarch64 wheel that's usable on 3.8.

This has already landed on `master` in 9c38b3be, which builds both `web` and `ml-api-cpu` for `linux/arm64`. It goes a bit further than this PR:

- Native arm64 runners (`runs-on: ubuntu-24.04-arm`) instead of QEMU emulation, which matters a lot for an image that pip-installs onnxruntime and downloads a 200 MB model.
- Digest-only push with a separate manifest-merge job, so one failing platform leg can't publish a half-populated manifest list.
- Per-platform GHA cache scopes at `mode=min` — four scopes share the repo's 10 GB cache, so a single shared `mode=max` scope across two platforms would evict itself.
- A runtime check (`.github/scripts/assert_inference_healthy.sh`) that boots the built image and waits for it to answer. That one is load-bearing for this image specifically: `lib/detection_model.py` swallows an `ImportError` from the ONNX backend, so a build that pip-installs cleanly can still produce an image whose gunicorn worker dies at boot. A build-only signal wouldn't catch it.

The reason you didn't see any of that: this PR targets `release`, which is currently 34 commits behind `master` and still carries the older revision of the workflow. We take PRs against `master` — `release` is the deploy branch and gets updated from `master`, so a hand-edit landing directly on it would have to be reconciled the next time the branches merge.

Closing as already-done. That said, arm64 images genuinely aren't published yet, because `master` hasn't reached `release` — that's on us, and separate from this PR. If you pull `master` and find the arm64 CPU image still doesn't run natively on your host, please do open an issue; that would be a real bug and exactly the kind of report we can't generate ourselves.

